PR #7, opened 2024-11-17
Feat/add linter

simon-the-shark:
so this sets up the linter, typecheck, prettiercheck ci workflow. Simple tests should work too, but if you need db in tests, then you'll probably need to set up it too (unless adonis has some built in mechanism for this shit, I have no clue)


.env.example
```diff
@@ -2,10 +2,11 @@ TZ=UTC
...
+APP_KEY=nice_keynice_keynice_keynice_keynice_keynice_key
 NODE_ENV=development
 DB_HOST=127.0.0.1
 DB_PORT=5432
 DB_USER=root
 DB_PASSWORD=root
-DB_DATABASE=app
\ No newline at end of file
+DB_DATABASE=app
+LOG_LEVEL=debug
```
czaja307, line 12: why there are 2 LOG_LEVELs?

simon-the-shark: good question

simon-the-shark: lets  do a quick check

czaja307: it seems to work ig

czaja307 (approved):

PR #26, opened 2024-11-23
Chore: Add migrations needed for multilayered map entities

Lennox2536:
Entities proposal for the multilayered map.

Diagram:
image

PR #31, opened 2024-11-25
Feat: Add adonis models for multilayered map entities

Lennox2536:
**Pull request with migrations**: https://github.com/Solvro/backend-topwr/pull/26


czaja307 (approved):
Well done 🔥

PR #50, opened 2024-12-02
fix: make all models use Luxon dateTime and fix typos

czaja307:
https://lucid.adonisjs.com/docs/models#date-columns

### DIFF
--- a/app/models/aed.ts
+++ b/app/models/aed.ts
@@ -14,10 +14,10 @@ export default class Aed extends BaseModel {
   declare longitude: number
 
   @column()
-  declare addresLine1: string
+  declare addressLine1: string
 
   @column()
-  declare addresLine2: string | null
+  declare addressLine2: string | null
 
   @column()
   declare photoUrl: string | null
--- a/app/models/bicycle_shower.ts
+++ b/app/models/bicycle_shower.ts
@@ -20,10 +20,10 @@ export default class BicycleShower extends BaseModel {
   declare longitude: number
 
   @column()
-  declare addresLine1: string
+  declare addressLine1: string
 
   @column()
-  declare addresLine2: string | null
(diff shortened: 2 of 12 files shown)


app/models/aed.ts
```diff
@@ -14,10 +14,10 @@ export default class Aed extends BaseModel {
   declare longitude: number
 
   @column()
-  declare addresLine1: string
+  declare addressLine1: string
 
   @column()
-  declare addresLine2: string | null
+  declare addressLine2: string | null
```
mini-bomba, line 20: This typo is also present in migrations :/

app/models/bicycle_shower.ts
```diff
@@ -20,10 +20,10 @@ export default class BicycleShower extends BaseModel {
   declare longitude: number
 
   @column()
-  declare addresLine1: string
+  declare addressLine1: string
 
   @column()
-  declare addresLine2: string | null
+  declare addressLine2: string | null
```
mini-bomba, line 26: Same issue as above

app/models/library.ts
```diff
@@ -16,10 +16,10 @@ export default class Library extends BaseModel {
   declare room: string | null
 
   @column()
-  declare addresLine1: string
+  declare addressLine1: string
 
   @column()
-  declare addresLine2: string | null
+  declare addressLine2: string | null
```
mini-bomba, line 22: Same issue once again

mini-bomba (changes requested):
Some of these changes will require a new migration to be created.

I've also found one more issue in the models:
- `department_link.ts` and `field_of_study.ts` have declared the foreign key columns `department_id` as strings instead of numbers

mini-bomba (approved):
I'm not impressed with the migration changes, but sure, :shipit:

PR #53, opened 2024-12-04
Feat/buildings routes

karolwolek:


app/models/building.ts
```diff
@@ -24,10 +24,10 @@ export default class Building extends BaseModel {
   @column()
   declare campusId: number
 
-  @column()
+  @column({ columnName: 'address_line1' })
```
czaja307, line 27: why did you do it?
not saying it is wrong, just asking to see idea behind it

start/routes.ts
```diff
@@ -9,15 +9,30 @@
...
-    router.get('files/:key', [FilesController, 'get'])
-    router.post('files', [FilesController, 'post'])
+    router.get('/:key', [FilesController, 'get'])
+    router.post('/', [FilesController, 'post'])
+  })
+  .prefix('api/v1/files')
+
+router
+  .group(() => {
+    // fetches campuses specifically
```
czaja307, line 29: i think that i'd be more RESTful if you used separate endpoint for campuses

app/controllers/buildings_controller.ts
```diff
@@ -0,0 +1,87 @@
+import Building from '#models/building'
+import Campus from '#models/campus'
+import { HttpContext } from '@adonisjs/core/http'
+export default class BuildingsController {
+  async getAll({ response }: HttpContext) {
```
czaja307, line 5: the function names are okay, but i'd change them to more adonis conventional
if you use `node ace make:controller [name] --api`, you will get the api template

the conventional names are
index - list of all records
store - create new record
show - get individual record
update - quite obvious :p
destroy - deletes record

what i'd suggest - create separate 2 separate controllers to avoid naming conflicts, and change function names to index and store (getByCampus is fine imo)

karolwolek: gotcha, will do

app/models/building.ts
karolwolek: so basically lucid is changing model field names to camel case in default. So addressLine1 will transfer to address_line_1 and thus its no valid in current migrations implementation (in diagrams and migrations its address_line1) so i thought less invasive is to set it here for address_line1 but i can change it other way around

start/routes.ts
```diff
@@ -9,15 +9,30 @@
...
+    router.get('/campuses/:id', [BuildingsController, 'getByCampus'])
```
czaja307, line 30: i think u used the wrong method here

app/models/building.ts
czaja307: makes sense, your version looks cleaner too
@Lennox2536 you might want to check this out, because you are using the same columns

database/seeders/building_seeder.ts
czaja307, line 1: seeders are a good idea for testing, but what u need to know is that they are executed in the alphabetical order and the campus_seeder needs to go before the buildings one, because of the dependencies

app/controllers/buildings_controller.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  async getById({ request, response }: HttpContext) {
+    try {
+      const id = request.param('id')
+      if (!id) {
+        return response.status(400).json({ message: 'Building ID is required' })
+      }
+      const building = await Building.find(id)
+      if (building) {
```
czaja307, line 26: when you are doing checks like this it's better to be specific (js logic can be wild sometimes) so i'd go with `building !== null` to explicitly show what you meant

app/controllers/buildings_controller.ts
```diff
@@ -0,0 +1,87 @@
...
+        return response.status(200).json(building.serialize())
```
czaja307, line 27: i'm quite sure that you don't need to run `serialize()` method, the adonis api setup should handle it for you

app/controllers/buildings_controller.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  async getByCampus({ request, response }: HttpContext) {
+    try {
+      const campusId = request.param('id')
+      if (!campusId) return response.status(400).json({ message: 'campus ID is required' })
+
+      const [campus, buildings] = await Promise.all([
+        Campus.find(campusId),
+        Building.query().where('campusId', campusId),
+      ])
```
mini-bomba, line 45: Couldn't we do the same thing using the `buildings` relationship on the `Campus` object?
(oh yeah also i've tried it and I think the relationship isn't defined correctly cuz it doesnt work, i have no idea why)

app/controllers/buildings_controller.ts
```diff
@@ -0,0 +1,87 @@
...
+      const campus = await Campus.find(id)
+      await campus?.load('buildings')
+      if (!campus) return response.status(404).json({ message: 'Campus not found' })
+
+      return response.status(200).json(campus.serialize())
+    } catch (error) {
+      return response.status(500).json({ message: 'Failed to fetch campus by Id' })
+    }
+  }
+}
```
mini-bomba, line 87: This function appears unused.

start/routes.ts
```diff
@@ -9,15 +9,30 @@
...
+    router.get('/campuses', [BuildingsController, 'getCampuses'])
+
+    // fetches buildings specifically
+    router.get('in-campus/:id', [BuildingsController, 'getByCampus'])
+    router.get('/:id', [BuildingsController, 'getById'])
+    router.get('/', [BuildingsController, 'getAll'])
   })
-  .prefix('api/v1')
+  .prefix('api/v1/buildings')
```
mini-bomba, line 38: 1. `/api/v1/buildings/in-campus/:id` is a duplicate of `/api/v1/buildings/campuses/:id`
2. `/api/v1/buildings/campuses` doesn't seem like the most logical path to get a listing of campuses. Maybe this should be exposed outside of the buildings group? (`/api/v1/campuses`)
   Or maybe it would make more sense to start with campuses? `/api/v1/campuses` for a list of campuses, `/api/v1/campuses/buildings` for a list of all buildings, `/api/v1/campuses/:id/buildings` for a list of buildings in a given campus?

mini-bomba (commented):
Seeing seeders getting implemented make me somewhat hopeful... that maybe, at some point, we'll be doing some unit tests...
But there are a few issues with the implementation at the moment:
1. The seeders are creating test data, but aren't constrained to run only in a development environment. You probably should constrain them to development and testing environments only. ([see the relevant docs](https://lucid.adonisjs.com/docs/seeders#environment-specific-seeders))
2. You've created two seeders, but the building seeder relies on data created by the campus seeder. I think that we should create seeders per schema and not per table, as that way we'll eliminate the dependencies between seeders. 
   Cross-seeder dependencies are an easy way to make mistakes, as ACE will run the files in alphabetical order. (just like migrations) If you don't make sure that the seeder you depend on ends up first when sorting, you'll end up with errors when trying to run them, as happened when I tried to run the seeders after creating the database from scratch.
   image

I also have some questions regarding the controller behaviour in edge cases:
1. Do we really need to return 404 when the query returns no results, but the query itself was valid? (specifically when the database has no building data, or when the provided campus ID was valid, but had no assigned buildings)
   IMO an empty list is already self-explanatory, 404 should be reserved for cases when the ID that the client provided was invalid.
2. Wrapping the controller methods in a try-catch seems a bit unnecessary to me. Adonis is already set up to catch errors and return 500 with the error message. The only thing this accomplishes is adding a generic context message. It also removes the log messages when such an error is thrown, which might make debugging harder.
   While the default adonis errors are quite bloated in development mode, they get reduced to basically the same thing as you're doing here with try/catches in production mode.
3. The API appears to return `campusId` fields of buildings as a string. I suspect this is because the foreign key column was defined as a `bigInteger` in the migrations. (cuz regular javascript integer numbers can only be 32bit, you need the bigint object to use 64bit numbers in js)
4. Is there a better way to validate the types of query parameters, one that isn't just manual checks in if statements?

app/controllers/buildings_controller.ts
mini-bomba: verified, true

karolwolek:
1. separated controllers into 2 different. Changed the methods mistake in router
2. rewrote endpoints
3. combined seeders together

app/controllers/buildings_controller.ts
mini-bomba, line 1: When I mentioned 404s and empty lists, I did not mean to swap the 404s and 400s. The codes returned were correct.

What I meant was that in cases where the user makes a request like `GET /api/v1/buildings/in-campus/2`, and a campus with that ID exists, but has no associated buildings, should we really be returning an error? I don't think it's necessarily a bad way to handle such an edge case, but since an empty array is also a valid, we might as well return an empty array with a 200 status code. The same goes for the list all buildings endpoint.

I think that returning an empty array makes it easier for the client to figure out what went wrong and display an appropriate message. The client might want to display an actual error message when the ID it requested was invalid, but all it might want to do if there are no associated buildings is just show some "No buildings" text.
If you return 404 for both cases, the client would likely need to parse the response message, which would break if someone changed the message. Also each endpoint might have a different message that means "valid ID, no associated objects".

And I'm not saying that you should change the status codes; they're already correct for their use case, that's not a good way to solve this issue.
I guess a good way to solve the error message parsing issue would be to include a `code` field in the error response, and have the codes be shared across all endpoints. We could have an `E_NO_RESULTS` code for the case where you would return an empty list, and `E_UNKNOWN_ID` for the case where the ID was invalid. (we probably should do these even if we decide to just return empty arrays in this case)

Anyway, not all of my review comments are meant to be "hey, fix this". Some of them (like this one) are meant to be an invitation into a discussion.

czaja307: What i'd propose
1. In the `index` functions go with `404`, because in normal scenario there should always be some buildings to display.
2. The same for the show - this seems to be pretty common usage of `404` to me.
3. For the `getByCampus` we can use the similar logic if we are sure that every campus should have at least one building. If that's not the case i'd personally go with `204`, which signalises that request was successful but there are no records.

what do you guys think about this?

mini-bomba: Adding a 204 case would make using the API more complex. (remember that 204 can never carry a response body, and even if you try, some clients will just ignore it)
IMO just returning an empty list with a 200 code would be the simplest.

karolwolek:
I had in mind @mini-bomba question about validation and i am also trying to implement custom validators for buildings and campuses using VineJS - adonis own validation framework.

karolwolek:
I have added required validators for user input in query params and route parameters. 
* The show methods will first ensure that the id is present in id column on specific table and only then attempt to fetch all columns. If id is not present vine.js will attach suitable error log in response.
* I added functionality to paginate response of lists of buildings and made it optional so if there are no query string params for pagination the api will try to send all records.
* I am not sure if changing BigInteger on foreigh key is a good approach, but i changed it for testing other functionalities so let me know about other ways for handling this
* Campuses routes have optional param for loading relational data from buildings by adding includeBuildings boolean param
* Buildings routes have optional param for loading relational data from campuses by adding includeCampus boolean param

database/migrations/1732381079405_create_buildings_table.ts
karolwolek, line 1: Not sure how to handle this

dawidlinek (commented):
Great job! I'm especially impressed with your engagement during the code review process. I have a suggestion: consider creating model hooks to handle recurring functionalities. These could include not only pagination but also filtering, sorting, and even relation loading.
For inspiration, you can refer to the https://github.com/Solvro/backend-cocktail-api/blob/main/app/controllers/ingredients_controller.ts. However, keep in mind that this can be optimized and done even better.
With such hooks in place, you can easily implement listing methods without relying on specific methods like getByCampus. This approach avoids the need to create a getByX method for every field.
Regarding the message field in your listing, it's a matter of preference. You can also  consider structuring the response as `{ message, ...buildings }` to keep it more concise or throw out the message field in successful requests all together.

app/controllers/buildings_controller.ts
```diff
@@ -1,70 +1,83 @@
...
-      return response
-        .status(500)
-        .json({ message: 'Failed to fetch buildings', error: error.message })
+  async index({ request, response }: HttpContext) {
+    const { page, limit, includeCampus } = await request.validateUsing(indexValidator)
+
+    const buildingsQuery = Building.query()
+    const message = 'List of buildings'
+
+    if (includeCampus) {
```
czaja307, line 15: in general i think that checking specifically `if (includeCampus !== undefined && includeCampus === true)` helps to avoid JS logic madness
image

app/controllers/buildings_controller.ts
```diff
@@ -1,70 +1,83 @@
...
+      buildingsQuery.preload('campus')
+    }
+
+    if (page) {
+      const buildings = await buildingsQuery.paginate(page, limit || 10)
```
czaja307, line 20: `||` works okay in this case, but in general when you are checking for undefined it is better to use `??` operator

app/controllers/buildings_controller.ts
```diff
@@ -1,70 +1,83 @@
...
-      }
-      return response.status(200).json(buildings)
-    } catch (error) {
-      return response
-        .status(500)
-        .json({ message: 'Failed to fetch buildings', error: error.message })
+  async index({ request, response }: HttpContext) {
+    const { page, limit, includeCampus } = await request.validateUsing(indexValidator)
+
+    const buildingsQuery = Building.query()
```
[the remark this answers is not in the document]
karolwolek: Correct me if i am wrong but the ``` Building.all() ``` promises  to return an array of ```Building``` model. 
The ```Building.query()``` returns a query builder instance, allowing you to chain methods to add additional query parameters.

app/controllers/buildings_controller.ts
czaja307: @mini-bomba global config coming soon™

app/controllers/buildings_controller.ts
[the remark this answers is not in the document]
czaja307: didn't see that, you both are correct

mini-bomba (approved):
Epic work! :fire: :shipit: 

I've just got a few last minor comments.

app/models/building.ts
```diff
@@ -3,13 +3,31 @@ import { DateTime } from "luxon";
 import { BaseModel, belongsTo, column, hasMany } from "@adonisjs/lucid/orm";
 import type { BelongsTo, HasMany } from "@adonisjs/lucid/types/relations";
 
+import { typedModel } from "../decorators/typed_model.js";
 import { BuildingIcon } from "../enums/building_icon.js";
+import { preloadRelations } from "../scopes/preload_helper.js";
+import { handleSearchQuery } from "../scopes/search_helper.js";
+import { handleSortQuery } from "../scopes/sort_helper.js";
```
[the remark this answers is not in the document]
karolwolek: thats a good point, will do

app/scopes/preload_helper.ts
```diff
@@ -0,0 +1,79 @@
...
+): ModelQueryBuilderContract<T> {
+  const relation = relationParts.shift();
+  if (relation === undefined) {
+    return query;
+  }
+
+  const relationDefinition = model.$relationsDefinitions.get(relation);
+  if (relationDefinition === undefined) {
+    return query;
+  }
```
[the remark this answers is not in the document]
karolwolek: > Is this `relationDefinition === undefined` a case where the string that someone put in was invalid/had an invalid relation? 

Yes, my first idea was to simply ignore it bc I thought any error status response is not needed. The lack of preloaded data
sent to user is self explanatory. Do you mean any particular way of logging it ?

app/scopes/search_helper.ts
```diff
@@ -0,0 +1,293 @@
...
+      query,
+      params: Partial<
+        Record<
+          keyof ModelAttributes<InstanceType<T>>,
+          undefined | string | string[] | FromTo
+        >
+      >,
+    ) => {
+      for (const param in params) {
+        const value: string | string[] | FromTo | undefined = params[param];
```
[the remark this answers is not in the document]
karolwolek: just tried it and typescript can't resolve value's type with current approach for params type in function parameters definition. I got unknown and linter gets angry about it in further functions

database/seeders/building_seeder.ts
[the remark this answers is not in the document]
karolwolek: i forgot about this comment and those seeders and i aggree. It was only for my specific tests cases. I don't even think it's anyway important in the main code anymore.

app/scopes/search_helper.ts
```diff
@@ -0,0 +1,293 @@
...
+        query = query.where(column.columnName, date.toJSDate());
+      } else {
+        valid = false;
+      }
+      break;
+    }
+    default: {
+      throw new NotImplementedException(
+        `Unsupported column type '${columnType}' in arrayTypeCheckHelper.`,
+      );
```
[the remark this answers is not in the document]
karolwolek: You left me thinking about this one and actually i could just apply here the same logic as for arrays: simply pass array with one value.
The code would change to sth like this 
```
      function handleDirectValue(
  query: ModelQueryBuilderContract<LucidModel, LucidRow>,
  column: ModelColumnOptions,
  value: string,
) {
  const columnType = extractType(column);
  const invalid = arrayTypeCheckHelper([value], columnType);
  if(invalid.length > 0) {
    throw new BadRequestException(
      `invalid filter value '${value}' ` +
        `for column: '${column.columnName}' ` +
        `of type: '${columnType}'`,
    );
  } else {
    if(columnType === "string") {
      query = query.whereILike(column.columnName, value);
    } else {
      query = query.where(column.columnName, value)
    }
    return query;
  }
}
```

app/controllers/buildings_controller.ts
```diff
@@ -0,0 +1,57 @@
...
+          "campusId",
+          "addressLine1",
+          "addressLine2",
+          "latitude",
+          "longitude",
+          "haveFood",
+          "createdAt",
+          "updatedAt",
+        ]),
+      );
```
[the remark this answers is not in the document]
karolwolek: > Or maybe the scope could just take the whole request and extract what it needs by itself?

Upside for manual pointing columns is to exclude and include just the columns you need. I intentionally excluded iconType column here since it does not have supported filtering type. This behaviour could be achieved by filtering object columns that have this type defined in meta property. But than again we loose the flexibility to exclude some other odd columns.

> Is there a way to do this without hard-coding every possible field in the controller?
```suggestion
        request.only([
          ...Object.keys(
            Building.$columnsDefinitions.entries().filter(([_, column]) => {
              const columnDef: ColumnDef = column;
              return columnDef.meta !== undefined
                ? columnDef.meta.type !== undefined
                : false;
            }),
          ),
        ]),
```
Could be done like so, dunno if I should even try to exclude columns with type other than allowed in this scope. Is it better than hardcoding? Hardcoded is more readable for sure, But suggested approach could  be more managable since you can get caught relying on word that "yes there's type for sure"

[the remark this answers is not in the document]
karolwolek: Looking at my suggestion i've got an idea that maybe it's actually a good approach to move this logic inside the scope as you said and leave an option to exclude some columns by user as a side parameters and only those would be hardcoded ?

app/scopes/preload_helper.ts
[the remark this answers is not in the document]
mini-bomba: I see that the controllers filter out only allowed relations from the request to pass down to the scope, so it's possible that someone might make a typo while creating those relation allowlists.
I was thinking that something as simple as a `console.log` (or another logging equivalent) with a message that an invalid relation was given would work.
This would mainly be useful when testing/debugging your own controllers locally.

app/scopes/search_helper.ts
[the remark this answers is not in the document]
mini-bomba: :fire::fire::fire:

app/controllers/buildings_controller.ts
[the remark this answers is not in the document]
mini-bomba: that's actually a good point, I didn't even notice that the iconType field was missing from that list.
I think that it's more likely that we'd want to exclude just a few fields rather than including just a few select ones.

app/controllers/campuses_controller.ts
```diff
@@ -0,0 +1,38 @@
...
+  async index({ request }: HttpContext) {
+    const campuses = await Campus.query().withScopes((scopes) => {
+      scopes.handleSearchQuery(
+        request.only(["id", "name", "createdAt", "updatedAt"]),
+      );
+      scopes.includeRelations(request.only(["buildings"]));
+      scopes.handleSortQuery(request.input("sort"));
+    });
+
+    return campuses;
```
czaja307, line 19: nitpick: you can return campuses inline, instead of creating the variable for it

app/decorators/typed_model.ts
```diff
@@ -0,0 +1,81 @@
...
+ *   - **Values**: Are types from the predefined `ColumnType` set (e.g., `'string'`, `'number'`, `'boolean'`, `'DateTime'`).
+ *
+ */
+export function typedModel<T extends LucidModel>(options: TypedModelOptions) {
+  return function (constructor: T): T {
+    const columns: Map<string, ModelColumnOptions> | undefined =
+      constructor.$columnsDefinitions;
+
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    if (columns === undefined) {
```
czaja307, line 60: question(linter): @Rei-x can you take a look at this line?

app/exceptions/not_implemented_exception.ts
czaja307, line 1: praise: custom exceptions look really cool

app/scopes/preload_helper.ts
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @template T - Type of the Lucid model.
+ * @param {T} model - The Lucid model to preload relations on.
+ * @returns {QueryScope} A scope function that takes a query and a relations object.
+ *
+ * @example
+ * const buildings = await Building.query()
+ *  .withScopes((scopes) => {
+ *    scopes.preloadRelations(request.only(["campus"]));
+ */
```
czaja307, line 24:
```suggestion
 *  });
```

app/scopes/preload_helper.ts
```diff
@@ -0,0 +1,79 @@
...
+          query = preloadSinglePath(query, model, relation.split("."));
+        }
+      }
+      return query;
+    },
+  );
+
+/**
+ * Helper function to work with recursively looaded chained relations
+ * - Base case [1] of function is the exhaustion of relationParts array (stack)
```
czaja307, line 44:
```suggestion
 * Helper function to work with recursively loaded chained relations
```

czaja307 (dismissed):
really good job 🔥

app/scopes/preload_helper.ts
czaja307: github messed it up, it's about typo in `looaded`

app/models/building.ts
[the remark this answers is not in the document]
karolwolek: I can't find a way to work around eslit screaming about those 
```ts
static preloadRelations = preloadRelations(Building);

static handleSearchQuery = handleSearchQuery(Building);

static handleSortQuery = handleSortQuery(Building);
```
when i defined imports as subpath. I will leave it for now. If you'll have time take a look if you want

czaja307 (approved):

app/decorators/typed_model.ts
mini-bomba: My guess is that because that field in the LucidModel is defined as just a Map, with no possibility of an undefined value, eslint can see that `columns` will never actually be undefined, even though it's defined here as possibly undefined.
image

PR #56, opened 2024-12-08
feat/organizations migrations

Batory12:
Please have a look if the link types are OK


mini-bomba (changes requested):
tl;dr: lots of minor spelling mistakes plus some actual issues and concerns

database/migrations/1733672331665_create_student_organisations_table.ts
```diff
@@ -0,0 +1,44 @@
...
+      table
+        .enum('organization_type', this.organization_type, {
+          useNative: true,
+          enumName: 'organization_type',
+          existingType: false,
+        })
+        .notNullable()
+      table.timestamp('created_at').notNullable()
+      table.timestamp('updated_at').notNullable()
+      table.foreign('department_id').references('departments.id').onDelete('CASCADE')
```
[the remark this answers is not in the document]
czaja307: yeah, the `RESTRICT` seems to be the safe option here

database/migrations/1733672331665_create_student_organisations_table.ts
```diff
@@ -0,0 +1,44 @@
+import { BaseSchema } from '@adonisjs/lucid/schema'
+
+export default class extends BaseSchema {
+  protected tableName = 'student_organisations'
+  protected sources = ['student_department', 'manual', 'pwr_active']
+  protected organization_type = [
```
czaja307, line 6: to keep everything neat and tidy, you should define your enums in `/app/enums` and import them here

database/migrations/1733688213674_create_student_organisation_links_table.ts
```diff
@@ -0,0 +1,44 @@
...
+  async up() {
+    this.schema.createTable(this.tableName, (table) => {
+      table.increments('id')
+      table.integer('student_organisation_id').unsigned().notNullable()
+      table
+        .enum('type', this.linkTypes, {
+          useNative: true,
+          enumName: 'organization_link_type',
+          existingType: false,
+        })
```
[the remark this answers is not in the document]
czaja307: mobile team asked that all link fields should share a single enum for link type

database/migrations/1733672331665_create_student_organisations_table.ts
mini-bomba: we've previously talked about how this wasn't a great idea if we ever decide to add enum values, as doing so will modify the behavior of already merged migration files without a direct change to them that would be visible in the git logs.

czaja307: okay, so just to get on the same page - you suggest to define enum used in models in `/app/enums` and the ones for migrations directly in the migration files?

mini-bomba: exactly, this way you can't possibly change the behavior of a migration file without editing it directly.

czaja307: alright, just keep reminding me if i keep mixing this up

Batory12:
OK, I fixed it. To the link_type enum: I just copied the values into the migration file, because that's how others do it, but what do you think maybe if the enum is shared,  the array should be shared as well, so we don't need to make modifications in every migration file?

czaja307:
> OK, I fixed it. To the link_type enum: I just copied the values into the migration file, because that's how others do it, but what do you think maybe if the enum is shared, the array should be shared as well, so we don't need to make modifications in every migration file?

you did it correct, i just got mixed up before
the fact that enum change forces us to change migration files is what we wanted to achieve (in general you shouldn't modify the migration files after you used them in production to avoid data loss - for now we can do that, because we do not have any data in db yet)
https://docs.solvro.pl/guides/adonis/#rollbacki-na-produkcji

database/migrations/1733687303819_create_student_organizations_student_organization_tags_table.ts
```diff
@@ -0,0 +1,27 @@
+import { BaseSchema } from "@adonisjs/lucid/schema";
+
+export default class extends BaseSchema {
+  protected tableName = "student_organizations_student_organization_tags";
```
mini-bomba, line 4: this table name is kinda ehhhhhhhh
but i guess it'll make it easier for adonis to detect it as a pivot table

mini-bomba (approved):
just one comment, but other than that :shipit:

database/migrations/1733687303819_create_student_organizations_student_organization_tags_table.ts
czaja307: i know it looks weird, but it was my idea to keep the naming convention

PR #66, opened 2024-12-16
Feat/adjust multilayered map schema

Lennox2536:
Applying requested changes to the multilayered map schema.

### DIFF
--- a/app/models/aed.ts
+++ b/app/models/aed.ts
@@ -16,7 +16,7 @@ export default class Aed extends BaseModel {
   declare longitude: number;
 
   @column()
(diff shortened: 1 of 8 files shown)


mini-bomba (commented):
don't forget to adjust the model definitions as well

czaja307:
as @mini-bomba said you'll need to update the models as well
migrations seems good, but if you can do the migrations in the original file (i know it's sinning but it'll keep the project cleaner) - just a suggestion, feel free to skip this one

czaja307 (approved):
LGTM, good job

PR #67, opened 2024-12-27
chore: define new subpath imports

mini-bomba:
This PR defines new subpath imports for the enums, scopes and decorators directories.


czaja307 (approved):
nice, LGTM

PR #72, opened 2025-01-05
fix(models): rename typedModel's meta field

mini-bomba:
The previous name ('type') was conflicting with lucid's internal field, which caused datetimes to not function on models decorated with the typedModel decorator.
I've renamed the field to 'declaredType' to avoid such conflicts as the issues arising from them are rather hard to debug and track down.

Discord discussion on this topic: https://discord.com/channels/687360174377533442/1325560731331985408


czaja307 (approved):
LGTM, express bugfix delivery 🔥

mini-bomba:
elo żelo

PR #75, opened 2025-01-09
feat: departments controllers

Mikuu-252:


app/controllers/departments_links_controller.ts
czaja307, line 1: i'm not sure if we need the specific endpoints just for links, imo using scopes on the departments model is just enough
it's discussable though, @mini-bomba what do u think about it?

app/models/department.ts
```diff
@@ -48,4 +52,10 @@ export default class Department extends BaseModel {
 
   @hasMany(() => DepartmentLink)
   declare departmentLink: HasMany<typeof DepartmentLink>;
+
+  static includeRelations = preloadRelations(Department);
```
czaja307, line 56: is there any reason why you decided to change the naming convention for this line?

database/seeders/department_seeder.ts
```diff
@@ -1,17 +1,99 @@
 import { BaseSeeder } from "@adonisjs/lucid/seeders";
 
 import Department from "#models/department";
+import DepartmentsLink from "#models/department_link";
+import FieldsOfStudy from "#models/field_of_study";
+
+import { LinkType } from "../../app/enums/link_type.js";
```
czaja307, line 7: you can use `#enums/link_type` instead for shorter notation

database/seeders/department_seeder.ts
```diff
@@ -1,17 +1,99 @@
...
   async run() {
-    await Department.create({
-      name: "Department 1",
-      addressLine1: "123 University Ave",
-      code: "123",
-      betterCode: "123",
-      gradientStart: "red",
-      gradientStop: "blue",
-    });
+    await Department.createMany([
```
czaja307, line 12: it's nice that you added such realistic data :D

app/controllers/departments_links_controller.ts
mini-bomba: I think fetching links through the departments endpoint as a relation should be enough.
There might be some argument for having an endpoint to get only the links of a department (without the other data), but imo getting a specific link by ID will never be useful.

app/controllers/departments_controller.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import Department from "#models/department";
+import { showValidator } from "#validators/show";
+
+export default class DepartmentsController {
+  async index({ request }: HttpContext) {
+    const departments = await Department.query().withScopes((scopes) => {
+      scopes.handleSearchQuery(request.only(["id", "createdAt", "updatedAt"]));
+      scopes.includeRelations(request.only(["departments"]));
+      scopes.handleSortQuery(request.input("sort"));
```
mini-bomba, line 11: - I think this endpoint should support more search query parameters than just the ID and dates. Ideally you'd only give the scope the few fields to ignore (such as the gradient colors) instead of a list of fields to allow
- `departments` is not a valid relation defined on the model. Look at the model definition to see what strings you should pass there - these should be the names of relations you defined there.
  - while at it you might as well move the list of valid relations into a readonly list on the class and reuse it in the `show` method without duplicating code

app/controllers/fields_of_studies_controller.ts
```diff
@@ -0,0 +1,36 @@
...
+import { paginationValidator } from "#validators/pagination";
+import { showValidator } from "#validators/show";
+
+export default class FieldsOfStudiesController {
+  async index({ request }: HttpContext) {
+    const { page, limit } = await request.validateUsing(paginationValidator);
+    const baseQuery = FieldsOfStudy.query().withScopes((scopes) => {
+      scopes.preloadRelations(request.only(["departments"]));
+      scopes.handleSortQuery(request.input("sort"));
+    });
```
mini-bomba, line 13: Similar issues as in the departments controller:
- `departments` is not a valid relation here, `department` is.
  - move the list of supported relations into a readonly class field to reduce code duplication
- surely we'd want to support filtering on this endpoint?

app/models/department.ts
mini-bomba, line 1: You must use the `@typedModel` decorator on the model definition here in order to use the filtering scope.

app/models/field_of_study.ts
mini-bomba, line 1: Same issue as the department model: use `@typedModel` to enable use of the filtering scope.

mini-bomba (changes requested):
<details>
<summary>minor mistakes</summary>

minor_mistake

</details>

app/controllers/departments_controller.ts
```diff
@@ -0,0 +1,40 @@
...
+          "betterCode",
+          "createdAt",
+          "updatedAt",
+        ]),
+      );
+      scopes.includeRelations(request.only(this.relations));
+      scopes.handleSortQuery(request.input("sort"));
+    });
+    return { data: departments };
+  }
```
mini-bomba, line 25: I think searching by address might be useful, which this code currently does not allow.
As I said in my previous review, it's quicker to only list fields that should not be searchable, as you'll likely have more searchable fields than those that are not.
In the case of this model, I think only the gradient colors and logo key should not be searchable.

Also this endpoint is missing the option for result pagination.
And the relation scope has a different name than in all other models.

app/controllers/departments_links_controller.ts
mini-bomba, line 1: Repeating comments from previous reviews again: An endpoint just for links isn't really necessary or useful - requesting them as a relation on a department is much more practical.

app/controllers/fields_of_studies_controller.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export default class FieldsOfStudiesController {
+  protected readonly relations = ["department"];
+
+  async index({ request }: HttpContext) {
+    const { page, limit } = await request.validateUsing(paginationValidator);
+    const baseQuery = FieldsOfStudy.query().withScopes((scopes) => {
+      scopes.preloadRelations(request.only(this.relations));
+      scopes.handleSortQuery(request.input("sort"));
+    });
```
mini-bomba, line 15: The scope for searching is missing.

mini-bomba (changes requested):
Nice work, but there are still some comments that weren't addressed.

app/controllers/departments_controller.ts
```diff
@@ -0,0 +1,47 @@
...
+    const { page, limit } = await request.validateUsing(paginationValidator);
+    const baseQuery = Department.query().withScopes((scopes) => {
+      scopes.handleSearchQuery(
+        request.except([
+          "logo",
+          "description",
+          "gradientStart",
+          "gradientStop",
+        ]),
+      );
```
mini-bomba, line 20:
```suggestion
      scopes.handleSearchQuery(
        request.qs(),
        "logo",
        "description",
        "gradientStart",
        "gradientStop",
      );
```
\+ prettier cuz i typed this out by hand

mini-bomba (commented):
Last round of changes?

czaja307 (dismissed):
@Mikuu-252 you have to resolve conflicts

mini-bomba (approved):
:fire: :shipit:

PR #80, opened 2025-01-19
feat/faq controllers and routes

Matistan:
pls review my pr
also, I didn't know whether to add seeders, if you want me to, I'll do it

### DIFF
--- a/app/controllers/guide_articles_controller.ts
+++ b/app/controllers/guide_articles_controller.ts
@@ -5,6 +5,7 @@ import { paginationValidator } from "#validators/pagination";
 import { showValidator } from "#validators/show";
 
 export default class GuideArticlesController {
+  protected readonly relations = ["guideAuthors", "guideQuestions"];
   /**
    * Display a list of resource
    */
@@ -12,7 +13,7 @@ export default class GuideArticlesController {
     const { page, limit } = await request.validateUsing(paginationValidator);
     const baseQuery = GuideArticle.query().withScopes((scopes) => {
       scopes.handleSearchQuery(request.qs(), "imagePath");
-      scopes.preloadRelations(request.only(["guideAuthors", "guideQuestions"]));
+      scopes.preloadRelations(request.only(this.relations));
       scopes.handleSortQuery(request.input("sort"));
     });
     let articles;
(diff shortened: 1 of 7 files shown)


mini-bomba:
Seeders aren't strictly required, but they make testing the endpoint much easier.

app/controllers/guide_articles_controller.ts
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Display a list of resource
+   */
+  async index({ request }: HttpContext) {
+    const { page, limit } = await request.validateUsing(paginationValidator);
+    const baseQuery = GuideArticle.query().withScopes((scopes) => {
+      scopes.handleSearchQuery(request.qs(), "imagePath");
+      scopes.preloadRelations(request.only(["guideAuthors", "guideQuestions"]));
+      scopes.handleSortQuery(request.input("sort"));
+    });
```
mini-bomba, line 17: you should move the list of possible relations into a readonly prop on the controller class and reuse it in both routes.
you look at some of the existing controllers and currently open PRs to see how to do that.
*this applies to all controllers added in this PR*

app/models/guide_article.ts
mini-bomba, line 1: You must use the `@typedModel` decorator on each model that has it's own controller in order for filtering to work in that controller

mini-bomba (changes requested):
<details>
<summary>minor issues</summary>

minor_issue

</details>

app/controllers/guide_articles_controller.ts
Matistan: ok, but even if a model has only 1 relation? some controllers leave `scopes.preloadRelations(request.only(["..."]));` when there is only 1 relation

mini-bomba: yeah, it's less likely to make a typo (or any other mistake involving these lists/strings) in one route and not the other this way

mini-bomba (approved):
:fire: job :shipit: 
the relations between guide articles and questions are broken because one of the columns is named differently than lucid would like it to be, but I'll fix it for you

mini-bomba (approved):

PR #81, opened 2025-01-19
feat: academic calendar controllers and routes

mateusz-plaska:
Controllers, routes, and seeder for academic calendars.


database/seeders/academic_calendar_and_holiday_and_day_swap_seeder.ts
mini-bomba, line 1: this file name could be *a bit* shorter.
`academic_calendar_seeder.ts` would be enough

start/routes.ts
```diff
@@ -88,3 +92,24 @@ router
...
+    router.get("/", [HolidaysController, "index"]);
+  })
+  .prefix("api/v1/holidays");
+
+router
+  .group(() => {
+    router.get("/:id", [DaySwapsController, "show"]);
+    router.get("/", [DaySwapsController, "index"]);
+  })
+  .prefix("api/v1/day-swaps");
```
mini-bomba, line 115: convert the route URLs from kebab-case to camel_case

app/models/academic_calendar.ts
mini-bomba, line 1: The relations defined on these models appear to be non-functional :/

<details>
<summary>screenshots of API results</summary>

image
image

</details>

mini-bomba (commented):
Good job, but there are some issues I encountered when testing this PR.

start/routes.ts
mateusz-plaska: Do you mean snake_case or camelCase?

mini-bomba: I meant snake_case, thanks for picking that up

app/models/academic_calendar.ts
mateusz-plaska: Right, I found the bug, but I have encountered the same problem in buildings controller. I'm gonna fix it also. And I can add libraries controller (this table is connected to buildings), I think that could be useful. One question that bothers me in  handleSearchQuery in buildings controller is why when I enter enum type in URL as a param it doesn't work (I mean nothing happens, no filtering)?

czaja307: opened new issue #82 for the second one

mini-bomba (changes requested):
Tested the routes (except libraries), relations appear to work - good job

<details>
<summary>but I did find some minor issues</summary>

caption

</details>

mini-bomba:
support for filtering by enum values just got merged into main, you should be able to properly define model fields that use enums after rebasing onto main.

mini-bomba (approved):
'ery noice :fire: :shipit:

PR #87, opened 2025-01-20
feat: about us models, migrations, and controllers

### LINKED ISSUE #84: feat: about us migrations

czaja307:
closes #84, #85, #86

i did it the way it is currently working on directus - there should be only 1 version of about us at any time (it's not the most advanced solution, but imo it'll stick)

skipped the query scopes, because i didn't think they were needed here

set the seeder to work in production environment - if you think that this is a bad approach, feel free to bring it up for discussion


app/controllers/about_us_controller.ts
mini-bomba, line 1: I don't really like the idea of having a whole table for a single record and not being able to limit it to one record only on the database level.
It just feels like it'd be too easy to screw up, someone could add a second record and wonder why it's not updating...
or worse, if `.query().first()` doesn't sort by any column, postgres will likely return a random/arbitrary row.

If we are limiting ourselves to only one about us version at any time, then there likely isn't any better way to handle this...
but we could consider allowing different about us versions for each language.
This way the ID key would be replaced with a language code as primary key, and we would avoid all of this confusion.
This would also allow each language version to have a different cover image. While at first thought this might be undesired behaviour, this would allow the image to contain localized text.

database/seeders/about_us_seeder.ts
mini-bomba, line 1: Allowing this seeder to run in production makes sense, because it isn't creating test data.

app/controllers/about_us_controller.ts
czaja307: good point, gonna change that

czaja307: @simon-the-shark take a look, do you want to keep it simple (leave pretty much the current directus version) or maybe extend the about_us section with multiple versions/languages support (we have a few ideas for that, but don't know if that's not an overkill)

czaja307:
@mini-bomba i've spoken to the Simon, and he says they don't really need multiple versions - i think we should use KISS principle and leave it as is

czaja307:
added constraint, we will have to remember to remove delete option for this in admin panel, because autoincrement will get stuck then

mini-bomba:
then disable autoincrement :brain:

mini-bomba:
could even set the id default to the correct id and hide the column from adonis

czaja307:
on it

app/models/about_us_general.ts
mini-bomba, line 1: `@typedModel` technically isn't needed in these models, but eh

database/migrations/1737364289031_create_about_us_generals_table.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  async down() {
+    this.schema.raw(
+      `ALTER TABLE ${this.tableName} DROP CONSTRAINT IF EXISTS single_row;`,
+    );
+
+    this.schema.dropTable(this.tableName);
+  }
+}
```
mini-bomba, line 28: Dropping the constraint shouldn't be necessary since we're about to drop the table anyway

database/seeders/about_us_seeder.ts
```diff
@@ -0,0 +1,93 @@
...
+        fs.copyFileSync(filePath, destination);
+      },
+    };
+
+    const result = await filesService.uploadFile(file as MultipartFile);
+
+    if (result instanceof Error) {
+      console.error(result);
+      return;
+    }
```
mini-bomba, line 59: I really dislike how this fileService works, with it returning an error instead of just throwing.

Also the error should probably be rethrown, since it's a critical error anyway.

and of course it throws when I try to use it, looks like it doesn't try to create the storage dir if it doesn't exist
image

app/models/about_us_general_link.ts
```diff
@@ -0,0 +1,30 @@
...
+@typedModel({
+  id: "number",
+  linkType: LinkType,
+  link: "string",
+  createdAt: "DateTime",
+  updatedAt: "DateTime",
+})
+export default class AboutUsGeneralLink extends BaseModel {
+  @column({ isPrimary: true, serializeAs: null })
+  declare id: number;
```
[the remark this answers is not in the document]
czaja307: id won't be shown when serialized to JSON

app/models/about_us_general.ts
czaja307: i know, but it may be useful in case we decide to extend it someday

app/models/about_us_general_link.ts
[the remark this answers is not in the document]
czaja307: yeah i meant to do it on the other model

database/seeders/about_us_seeder.ts
czaja307: created new issue #96

app/admin/resources/about_us.ts
```diff
@@ -0,0 +1,29 @@
...
+  resource: new LucidResource(AboutUsGeneral, "postgres"),
+  options: {
+    properties: {
+      id: {
+        isVisible: false,
+      },
+      ...readOnlyTimestamps,
+    },
+  },
+};
```
mini-bomba, line 19: Actions such as `new`, `delete` and `bulkDelete` should be disabled for this resource, as there can only ever be one of them, it should never be deleted and there's a production seeder that creates the first instance automatically.
```diff
 export const aboutUsResource = {
   resource: new LucidResource(AboutUsGeneral, "postgres"),
   options: {
     properties: {
       id: {
         isVisible: false,
       },
       ...readOnlyTimestamps,
     },
+    actions: {
+      new: {
+        isAccessible: false,
+        isVisible: false,
+      },
+      delete: {
+        isAccessible: false,
+        isVisible: false,
+      },
+      bulkDelete: {
+        isAccessible: false,
+        isVisible: false,
+      },
+    },
   },
 };
```

database/seeders/about_us_seeder.ts
```diff
@@ -0,0 +1,100 @@
...
+          "..",
+          "storage",
+          key,
+        );
+
+        // Ensure the /storage directory exists
+        const storageDir = path.dirname(destination);
+        if (!fs.existsSync(storageDir)) {
+          fs.mkdirSync(storageDir, { recursive: true });
+        }
```
mini-bomba, line 55: `/config/drive.ts` is probably a better place for this check

mini-bomba (commented):
final comments, finally

mini-bomba (approved):
:fire: :shipit:

PR #89, opened 2025-01-23
fix: actually implement searching by enum values

### LINKED ISSUE #82: fix: enums in query scopes not working properly
Right, I found the bug, but I have encountered the same problem in buildings controller. I'm gonna fix it also. And I can add libraries controller (this table is connected to buildings), I think that could be useful. One question that bothers me in  handleSearchQuery in buildings controller is why when I enter enum type in URL as a param it doesn't work (I mean nothing happens, no filtering)?

_Originally posted by @mateusz-plaska in https://github.com/Solvro/backend-topwr/pull/81#discussion_r1921844431_

mini-bomba:
fixes #82 

please let me know if I've unknowingly (or knowingly) committed something that could be considered a programming war crime

but um, yeah, i've actually implemented proper enum support for typed models and search scopes
to use, simply pass in the enum type to the `@typedModel` decorator
under the hood, it will assign the column an "enum" type and embed allowed enum values in a new meta prop

### DIFF
--- a/app/decorators/typed_model.ts
+++ b/app/decorators/typed_model.ts
@@ -4,13 +4,16 @@ import {
   ModelColumnOptions,
 } from "@adonisjs/lucid/types/model";
 
-export type ColumnType = "string" | "number" | "boolean" | "DateTime";
+type SharedColumnTypes = "string" | "number" | "boolean" | "DateTime";
+export type ColumnType = SharedColumnTypes | "enum";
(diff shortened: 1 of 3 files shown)


karolwolek:
Not a war crime but a great way to solve this. I didn't know how exactly string enum is treated at runtime (turns out it's a simple js object)

🔥 🔥 🔥

czaja307 (approved):
looks good, doesn't look like a war crime

PR #98, opened 2025-01-28
feat: create faq script to migrate the data

Matistan:
## PR notes:
- some articles had null descriptions, but in my model this field is notNullable, so they are now just empty
- I've made some duplicate questions due to some belonging to few articles
- Previously, createdAt and updatedAt were associated with the questions, not articles. I’ve changed the logic to now check for the latest createdAt and updatedAt dates across all questions for each article, and then assign these dates to the article.
- I've checked that the script works by connecting and testing on local database

I kindly ask for a review

----

> [!IMPORTANT]
> Adds `faq.ts` script to migrate FAQ data, handling null descriptions, duplicate questions, and updating article timestamps based on questions.
> 
>   - **Behavior**:
>     - New script `faq.ts` added to migrate FAQ data.
>     - Handles null `description` by setting it to an empty string in `GuideArticle.create()`.
>     - Duplicates questions if they belong to multiple articles.
>     - Updates `createdAt` and `updatedAt` for articles based on the latest question timestamps.
>   - **Error Handling**:
>     - Logs warnings for missing questions in `pivotTableResult` and unreferenced questions in `questionsResult`.
>     - Catches and logs errors during image upload in `uploadImage()`.
>   - **Misc**:
>     - Verifies script functionality by testing on a local database.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=Solvro%2Fbackend-topwr&utm_source=github&utm_medium=referral)<sup> for e4bb58e031b28a93017c3330d20afdea955c952f. It will automatically update as commits are pushed.</sup>


scripts/faq_script.ts
```diff
@@ -0,0 +1,124 @@
...
+      console.error("Question not found");
+      return;
+    }
+
+    await GuideQuestion.create({
+      title: question.question,
+      answer: question.answer,
+      articleId: pivot.FAQ_Types_id,
+    });
+  }
```
mini-bomba, line 121: It appears that this might implicitly skip importing some questions that aren't linked to a guide article. Could we emit warnings for those?

mini-bomba (changes requested):
as i always say, just some minor issues, nothing major :)

scripts/faq_script.ts
```diff
@@ -0,0 +1,124 @@
...
+      !pivotTableResponse.ok
+    ) {
+      console.error("Error fetching data");
+    }
+  } catch {
+    console.error("Error fetching data");
+    return;
+  }
+
+  console.log("Fetching data...");
```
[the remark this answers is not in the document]
czaja307: the linter allows you to use the `console.log` in the `/scripts` directory

scripts/faq_script.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    await GuideArticle.create({
+      id: article.id,
+      title: article.name,
+      shortDesc: article.short_description,
+      description: article.description ?? "",
+      imagePath: article.cover,
+      createdAt,
+      updatedAt,
+    });
```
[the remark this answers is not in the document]
Matistan: Could you tell me how can I scrap these images and add them to the new storage service?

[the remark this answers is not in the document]
mini-bomba: for downloading from the existing admin panel: I'd probably just fetch the image URL and get the blob with the image contents. I don't know what the image URL scheme is though. (cc @czaja307)

for uploading to the new storage service: take a look at [how the about us seeder does it](https://github.com/Solvro/backend-topwr/blob/main/database/seeders/about_us_seeder.ts#L39-L63). I haven't looked into this myself yet, so there might be an easier way to create a MultipartFile from a blob (or similar) than just impersonating it with a plain object. In the worst case you'll just have to create a moveToDisk function that literally just writes the blob to disk.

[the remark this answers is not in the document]
czaja307: `https://admin.topwr.solvro.pl/assets/`

mini-bomba (changes requested):

Matistan:
is everything correct now?

scripts/faq_script.ts
```diff
@@ -0,0 +1,183 @@
...
+        continue;
+      }
+
+      if (DateTime.fromISO(question.date_created) < createdAt) {
+        createdAt = DateTime.fromISO(question.date_created);
+      }
+
+      if (DateTime.fromISO(question.date_updated) > updatedAt) {
+        updatedAt = DateTime.fromISO(question.date_updated);
+      }
```
mini-bomba, line 136: minor nitpick: it might be wise to compute the `DateTime` once per object and save it to a temp variable, rather than recomputing it again if the condition is satisfied.

mini-bomba (changes requested):
sorry for the delay in reviews (sesja moment)
this week-long break made me spot some additional issues in the PR
(minor issues of course, as always)

scripts/faq_script.ts
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    const imagePath = await uploadImage(
+      `https://admin.topwr.solvro.pl/items/assets${article.cover}`,
```
czaja307, line 140: this endpoint does not seem to be correct 
```suggestion
      `https://admin.topwr.solvro.pl/assets/${article.cover}`,
```

Matistan:
pls check if everything is correct now, or did I miss something

scripts/faq_script.ts
```diff
@@ -0,0 +1,168 @@
...
+  const filesService = new FilesService();
+
+  const fetchedImage = await fetch(imageUrl);
+  if (!fetchedImage.ok) {
+    throw new Error(
+      `Failed to download image. HTTP status: ${fetchedImage.status}`,
+    );
+  }
+
+  const extension = fetchedImage.headers.get("content-type")?.split("/")[1];
```
mini-bomba, line 53: my concern about the way the file extension is derived still wasn't addressed.
I've found out that there's an endpoint for file metadata in directus and I myself just extract the extension from the on-disk filename that directus reports.
The endpoint URL is `https://admin.topwr.solvro.pl/files/:id`.

mini-bomba (changes requested):
Apart from that one file extension related concern, consider moving the script to the new scraper "framework" - this should be as simple as moving `faq_script.ts` to `database/scrapers` and wrapping the main function in a class.
If you don't feel like doing that, then I can do it for you once you address that one last review comment.

simon-the-shark:
@ellipsis review this

scripts/faq_script.ts
```diff
@@ -0,0 +1,168 @@
...
+        `Pivot references missing question (ID=${pivot.FAQ_id}) for article ID=${pivot.FAQ_Types_id}. ` +
+          `This may be a data inconsistency in the source. Skipping this question...`,
+      );
+      continue;
+    }
+
+    await GuideQuestion.create({
+      title: question.question,
+      answer: question.answer,
+      articleId: pivot.FAQ_Types_id ?? 0,
```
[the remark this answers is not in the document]
Matistan: Ik, but without it, there's ts error: Type `number | null` is not assignable to type `number | undefined`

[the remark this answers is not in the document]
mini-bomba: setting this to 0 will make it throw an error at runtime instead

database/scrapers/faq_script.ts
```diff
@@ -0,0 +1,190 @@
...
+  const extension = imageMetadata.data.type.split("/")[1];
+
+  const stream = Readable.fromWeb(fetchedImage.body as ReadableStream);
+
+  try {
+    return await filesService.uploadStream(stream, extension);
+  } catch (error) {
+    throw new Error(`Failed to upload image. ${error}`);
+  }
+}
```
mini-bomba, line 79: "Failed to upload image" adds no extra context.
Any error from `FileService.upload*()` will already have the top message say "Couldn't upload the file".
Either include what image you were unable to migrate, or don't catch the error.
If you decide to include more context, then please set the original error as the cause instead of putting only it's message in the error, as currently you're losing basically all error information. (see the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error) for more info/examples on how to do this)

database/scrapers/faq_script.ts
```diff
@@ -0,0 +1,190 @@
...
+
+export default class FaqScript {
+  static async run() {
+    await faqScript();
+  }
+}
```
mini-bomba, line 85: yeah this won't work
take a look at [other scraper modules](https://github.com/Solvro/backend-topwr/tree/main/database/scrapers) to see how to properly register a scraper module.

the issues:
- the class must extend `BaseScraperModule` (imported from `#commands/db_scrape`
- the class must have static `name` and `description` properties (+ optional `taskTitle`)
- `run()` cannot be static
- move the `faqScript()` function contents into the `run()` method
  - use `this.logger` instead of `logger`

database/scrapers/faq_script.ts
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  if (!fetchedImageMetadata.ok) {
+    throw new Error(
+      `Failed to fetch the image metadata. HTTP status: ${fetchedImageMetadata.status}`,
+    );
+  }
+
+  const imageMetadata = (await fetchedImageMetadata.json()) as ImageMetadata;
+  const extension = imageMetadata.data.type.split("/")[1];
```
mini-bomba, line 70: this did not solve my previous concerns: you're basically still doing the same thing, you're still using the last part of the mime type which is not guaranteed to be a valid file extension.
extract the extension from one of the filenames that appear in the metadata instead

database/scrapers/faq_script.ts
```diff
@@ -0,0 +1,190 @@
...
+          `This may be a data inconsistency in the source. Skipping this question...`,
+      );
+      continue;
+    }
+
+    await GuideQuestion.create({
+      title: question.question,
+      answer: question.answer,
+      articleId: pivot.FAQ_Types_id ?? undefined,
+    });
```
mini-bomba, line 178: if TS does not understand that .filter() removed the possibility of pivot.FAQ_Types_id being null, then just use a regular if statement + continue in the loop instead of .filter(), instead of doing this funky shit

image

might even be faster since it's not allocating another temporary array... but performance is kinda irrelevant in the context of these scrapers/migration scripts
as long as it completes in a reasonable amount of time, it's fine

mini-bomba (changes requested):
❌ Changes requested. Reviewed everything up to 27026d5 in like an hour or something

database/scrapers/faq_script.ts
```diff
@@ -67,124 +67,143 @@ async function uploadImage(imageUrl: string) {
...
+    if (imageMetadata.data.filename_download !== null) {
+      if (imageMetadata.data.filename_download.split(".").length > 1) {
+        extension = imageMetadata.data.filename_download.split(".").pop();
+      }
+    }
+  }
+
+  if (extension === undefined || extension === "") {
+    throw new Error("Failed to determine image extension");
+  }
```
mini-bomba, line 86: the extension being undefined or empty is not a critical error, the `uploadStream` function will use a `.bin` extension in these cases instead.
you can log some message in this case if you want.

database/scrapers/faq_script.ts
```diff
@@ -67,124 +67,143 @@ async function uploadImage(imageUrl: string) {
   }
 
   const imageMetadata = (await fetchedImageMetadata.json()) as ImageMetadata;
-  const extension = imageMetadata.data.type.split("/")[1];
+  let extension: string | undefined = "";
+  if (imageMetadata.data.filename_disk !== null) {
+    if (imageMetadata.data.filename_disk.split(".").length > 1) {
+      extension = imageMetadata.data.filename_disk.split(".").pop();
+    }
```
mini-bomba, line 74: minor: you're running split twice, consider not doing that

database/scrapers/faq_script.ts
```diff
@@ -0,0 +1,209 @@
...
+    throw new Error("Failed to determine image extension");
+  }
+
+  const stream = Readable.fromWeb(fetchedImage.body as ReadableStream);
+
+  try {
+    return await filesService.uploadStream(stream, extension);
+  } catch (error) {
+    throw new Error(`Failed to upload image.`, { cause: error });
+  }
```
mini-bomba, line 94: comment from previous review still unaddressed:
"Failed to upload image" adds no context.
If you can't add any context here, consider not catching the error and trying to add context one level higher

mini-bomba (commented):
the minor issues never end

database/scrapers/faq_script.ts
```diff
@@ -0,0 +1,218 @@
...
+        extension = filenameDownloadParts.pop();
+      }
+    }
+  }
+
+  if (extension === undefined || extension === "") {
+    logger.warning(
+      "Failed to determine image extension, using a .bin instead.",
+    );
+  }
```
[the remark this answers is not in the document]
Matistan: I would rather have a real person review this one, if you don't mind 😁

Matistan:
I really wish that this was the last portion of minor issues...

database/scrapers/faq_script.ts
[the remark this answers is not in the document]
mini-bomba: unnecessary suggestion; thankfully it won't be responding automatically anymore for new prs

mini-bomba (approved):
:fire: :shipit:

PR #101, opened 2025-02-01
Refactor: improve error handling

kamilfedio:
I notice that the creation of `storage` has been already implemented - so I have ommited this

- created custom errors
- removed previous solutions


app/services/files_service.ts
```diff
@@ -3,32 +3,35 @@ import { randomUUID } from "node:crypto";
...
-  async deleteFile(key: string): Promise<void | Error> {
+  async deleteFile(key: string): Promise<void> {
     // Delete file from storage
-    try {
-      await drive.use().delete(key);
-    } catch (error) {
-      return error;
-    }
+    await drive.use().delete(key);
   }
```
mini-bomba, line 36: surprised we aren't wrapping the errors here in a custom error class

app/exceptions/file_service_errors.ts
```diff
@@ -0,0 +1,15 @@
...
+  }
+}
+
+export class FileServiceFileReadError extends Error {
+  constructor(message: string, stack: string | undefined) {
+    super(message);
+    this.name = "FileServiceFileReadError";
+    this.stack = stack;
+  }
+}
```
mini-bomba, line 15: maybe we should have a common base class for these exceptions? `FileServiceError` as base, and specific errors like `FileServiceFileReadError` as an extension of the base class?

mini-bomba (commented):
:fire: :fire: :fire: hell yeah, actually sane error handling, as commonly accepted in this programming language

looks good overall, but i've found a potential inconsistency and one more minor way to improve this

app/services/files_service.ts
kamilfedio: https://flydrive.dev/docs/disk_api#delete

This doesn't throw an errow by docs

app/exceptions/file_service_errors.ts
kamilfedio: done

app/services/files_service.ts
mini-bomba: :thinking: even on something as stupid as a permission issue?
time to test i guess, give me a moment
I hope the docs are wrong, because I'd hate for it to just be a silent error

kamilfedio: yeah, I thought the same - just flydrive things

mini-bomba: Verified: if I properly break the permissions on the storage directory (which took me 3 tries), the deleteFile() method will indeed throw an error as I expected.
The docs are wrong.
image

app/exceptions/file_service_errors.ts
```diff
@@ -0,0 +1,21 @@
+class FileServiceError extends Error {
+  constructor(message: string, stack: string | undefined) {
+    super(message);
+    this.name = "FileServiceFileUploadError";
```
mini-bomba, line 4: you probably shouldn't be setting this name in the base class

kamilfedio: my miss

app/exceptions/file_service_errors.ts
mini-bomba, line 1: I think we should be setting the `cause` field instead of just overwriting the `stack` property. (unless adonis doesn't make it visible in the logs or responses [but it's a standard field, so it should])
Looks like this should be possible to do just by passing in the original error as an additional parameter to the superclass' constructor

mini-bomba: and that way the message of the error could be a bit more specific (even if it's just "couldn't create the file", it should be a bit friendlier and will give away less crucial details on prod, while hopefully still logging the full error in the logs)

PR #103, opened 2025-02-12
feat(fileService): add more methods for uploading files

mini-bomba:
This PR adds methods such as `uploadStream` and `uploadFromMemory`, which aim to replace the existing hacks of impersonating the `MultipartFile` interface to upload custom files.

Now, you can simply call one of these functions with an appropriate input type, which should be easier to construct than `MultipartFile`.

### DIFF
--- a/app/controllers/files_controller.ts
+++ b/app/controllers/files_controller.ts
@@ -11,7 +11,7 @@ export default class FilesController {
       return response.badRequest("No file provided");
     }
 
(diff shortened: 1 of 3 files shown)


czaja307 (approved):
LGTM, it'll certainly make the life easier

just one suggestion - what do you think about converting the comments to DocStrings?
imo it could be helpful, but feel free to merge it as is

czaja307 (approved):
🔥 :shipit:

PR #109, opened 2025-02-17
fix(scraper): console.error() module errors instead of using cliui

mini-bomba:
cliui doesn't print error causes, which produces worse errors
so let's just console.error() the error out instead

before
image

after
image


czaja307 (approved):
LGTM

PR #111, opened 2025-02-20
feat: buildings and campuses script

karolwolek:
> [!IMPORTANT]
> Adds a script for scraping building and campus data, updates models and migrations, and enhances existing functionality.
> 
>   - **New Features**:
>     - Adds `BuildingsScraper` in `database/scrapers/buildings.ts` to scrape building and campus data from Directus and local JSON.
>     - Introduces `campuses.json` for local campus data.
>   - **Models**:
>     - Updates `Building` and `Campus` models to include new fields `externalDigitalGuideMode` and `externalDigitalGuideIdOrURL`.
>     - Adds `serializeExtras` to `Contributor` and `Role` models.
>   - **Migrations**:
>     - Updates `1732380915866_create_campuses_table.ts` and `1732381079405_create_buildings_table.ts` to make `cover` nullable and add new fields.
>   - **Controllers**:
>     - Modifies `StudentOrganizationsController` to handle pagination and sorting more robustly.
>   - **Decorators**:
>     - Enhances `typedModel` decorator in `typed_model.ts` to support enums.
>   - **Misc**:
>     - Fixes typo in `preload_helper.ts` documentation.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=Solvro%2Fbackend-topwr&utm_source=github&utm_medium=referral)<sup> for 3318b36fbc484154dc13b2fe142e566ba63f328d. It will automatically update as commits are pushed.</sup>


database/migrations/1732381079405_create_buildings_table.ts
```diff
@@ -30,6 +30,28 @@ export default class extends BaseSchema {
...
+        )
+        OR
+        (
+          external_digital_guide_mode IS NULL
+          AND external_digital_guide_id_or_url IS NULL
+        )
+      `,
+        [],
+        "digital_guide_format_check",
+      );
```
czaja307, line 54: praise: good thing that you made such explicit checks...
but i wish we didn't have to deal with such bs external APIs :sob:

app/scopes/preload_helper.ts
czaja307, line 1: nitpick: add the `@template T` line in the `preloadSinglePath` function

database/scrapers/building_scrapper.ts
czaja307, line 1: issue (blocking): Przewo would not approve these exceptions 

The way you are using exceptions here does not make sense to me. You throw them just to catch them later in the function. The better approach would be to log the edge case where you need it, then simply return.

czaja307 (changes requested):
left a few comment, feel free to discuss as always

overall solid job :fire:

app/models/contributor.ts
```diff
@@ -55,4 +55,6 @@ export default class Contributor extends BaseModel {
   static preloadRelations = preloadRelations(Contributor);
   static handleSearchQuery = handleSearchQuery(Contributor);
   static handleSortQuery = handleSortQuery(Contributor);
+
+  serializeExtras = true;
```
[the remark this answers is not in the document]
karolwolek: looking for an answer how to get pivot tables fetched alongside related tables i found that they are actually fetched but not serialized for json response. They are hidden in `.$extras` on model instance
here it is described where pivot columns are stored: https://lucid.adonisjs.com/docs/relationships#many-to-many-pivot-columns
here it is described how to serialize them: https://lucid.adonisjs.com/docs/serializing-models#serializing-extras-object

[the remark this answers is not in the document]
czaja307: makes sense, thanks

database/scrapers/building_scrapper.ts
```diff
@@ -0,0 +1,158 @@
...
+        throw new Error("failed reading ./assets/campuses.json");
+      }
+      const campusesMap = new Map<string, Campus>();
+      for (const data of campusesData.data) {
+        const campus = await Campus.create(data);
+        data.buildings.forEach((building) => campusesMap.set(building, campus));
+      }
+      this.logger.info("campuses created!");
+
+      this.logger.info("starting fetching buildings...");
```
mini-bomba, line 60: minor nitpick: you can use task.update() for such progress updates

database/scrapers/building_scrapper.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  async run(task: TaskHandle) {
+    try {
+      this.logger.info("starting reading campuses file...");
+      const campusesData = JSON.parse(
+        await fs.readFile("./assets/campuses.json", { encoding: "utf-8" }),
+      ) as CampusesData | undefined;
+      if (campusesData === undefined) {
+        throw new Error("failed reading ./assets/campuses.json");
+      }
```
mini-bomba, line 52: MDN docs for `JSON.parse` don't mention `undefined` as a possibility, but other JS types are possible.
I'd check if the result is an `object` and throw an error if it's not (mentioning what the actual type was)

database/scrapers/building_scrapper.ts
```diff
@@ -0,0 +1,158 @@
...
+      const campusesMap = new Map<string, Campus>();
+      for (const data of campusesData.data) {
+        const campus = await Campus.create(data);
+        data.buildings.forEach((building) => campusesMap.set(building, campus));
+      }
```
mini-bomba, line 57: use of `forEach` is generally discouraged.
if you want to avoid using for loops, you can pass the `Map` constructor a list of `[key, value]` tuples to initialize the map with.

database/scrapers/building_scrapper.ts
```diff
@@ -0,0 +1,158 @@
...
+      this.logger.info("campuses created!");
+
+      this.logger.info("starting fetching buildings...");
+      const response = await fetch(buildingsPath);
+      if (!response.ok) {
+        throw new Error(
+          `Error: ${response.status} ` + `cause: ${response.statusText}`,
+        );
+      }
```
mini-bomba, line 66: This error tells me nothing.
It'll output something like `Error: 404 cause: Not Found`.
Please include information about what led to this error (e.g. what you were trying to fetch)
A better error message here would be something like `Failed to fetch the buildings - got response status code ${response.status} ${response.statusText}`.

database/scrapers/building_scrapper.ts
```diff
@@ -0,0 +1,158 @@
...
+      const buildingsBody = (await response.json()) as
+        | { data: BuildingsData[] }
+        | undefined;
+      if (buildingsBody === undefined) {
+        throw new Error("failed fetching buildings: empty body");
+      }
```
mini-bomba, line 72: While I appreciate actually checking the result type, this check is faulty.
`undefined` is not representable in JSON and as such it will never be returned from this function. (it's literally the only value that you're guaranteed to never get here)
If you want to properly check the type, do a `typeof` check and ensure it is an `object`.

database/scrapers/building_scrapper.ts
```diff
@@ -0,0 +1,158 @@
...
+      this.logger.error(`failed to upload the files: ${err}`);
+      return "";
+    }
+  }
+}
+
+function resolveDate(dateString: string): DateTime<true> {
+  const date = DateTime.fromISO(dateString);
+  return date.isValid ? date : DateTime.local();
+}
```
mini-bomba, line 158: minor nitpick: `DateTime.local()` is the same as `DateTime.now()`, according to the docs.

database/scrapers/building_scrapper.ts
```diff
@@ -0,0 +1,158 @@
...
+          this.logger.error(
+            `No campus assigned to building ${buildingEntry.identifier}`,
+          );
+        } else {
+          // update campuses
+          campus.cover = buildingEntry.cover;
+          buildingEntry.campusId = campus.id;
+          await campus.save();
+        }
+      }
```
mini-bomba, line 109: just a minor code style nitpick: consider using `continue` in the if statement, so that you can avoid using an else branch.
... or maybe you should throw instead? If you continue without assigning the campusId, the database operating will just fail due to the foreign key constraint not being met.

database/scrapers/building_scrapper.ts
```diff
@@ -0,0 +1,158 @@
...
+      await Building.createMany(formattedBuildingData);
+      this.logger.info("Buildings created !");
+      task.update(`finished running ${this.name} scrapper`);
```
mini-bomba, line 112: minor nitpick: you don't need an update at the end of the function, as a "Done" success message will be printed as soon as you return from the function.

database/scrapers/building_scrapper.ts
```diff
@@ -0,0 +1,158 @@
...
+    }
+  }
+
+  private async uploadAndGetKey(data: BuildingsData): Promise<string> {
+    try {
+      const imageKey = data.cover;
+      if (imageKey === null) {
+        this.logger.warning(`no image for building: [${data.id}]`);
+        return "";
+      }
```
mini-bomba, line 125: don't return fake values, either throw an error, or make `null` a valid value.

database/scrapers/building_scrapper.ts
```diff
@@ -0,0 +1,158 @@
...
+        );
+      }
+      let mediaType = "";
+      for (const [header, value] of response.headers.entries()) {
+        if (header === "content-type") {
+          if (value.startsWith("image")) {
+            mediaType = value.split("/").pop() ?? "";
+          }
+        }
+      }
```
mini-bomba, line 143: As I noted in another PR, extracting the last part of the mime type is not a very robust mechanism for deriving the file extension.
Consider extracting the correct extension from directus' file metadata endpoint instead (`https://admin.topwr.solvro.pl/files/:id`)

mini-bomba (changes requested):
The way error handling was done here is probably my biggest concern right now.
Please make sure errors are clear and include (and preserve) as much context as possible, I don't want a giant mess to debug later.

database/scrapers/building_scrapper.ts
czaja307: are you sure, though? - `.forEach` with await is definitely bad, but i think that `for ... of` is okay
image
https://gist.github.com/joeytwiddle/37d2085425c049629b80956d3c618971

database/migrations/1732381079405_create_buildings_table.ts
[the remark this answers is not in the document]
czaja307: leave it as is imo, we will have to nuke the db anyway

database/scrapers/building_scrapper.ts
mini-bomba: it's not as disastrous as trying to use an async function with foreach, but I've heard that it's generally slower than just using a for loop [citation needed]
but generally I'd get using `.forEach` at the end of a long chain of functional array ops (`.map`, `.filter`, etc.) as I feel there's some potential for cleaner code there, but here it's being ran directly on an object property, which can be easily replaced with a for loop which imo would be cleaner in this case.
besides, if we're already writing some functional-style code, we might as well go all in
```ts
const campusesMap = new Map(
  await Promise.all(
    campusesData.data.map(async data => {
      const campus = await Campus.create(data);
      return data.buildings.map(building => [building, campus]);
    })
  ).then(r => r.flat(1))
)
```

karolwolek:
I changed the way error handling was done, but i think i will still find it usefull to use the semaphore lib @mini-bomba brought in pr #113. Especially in the fetch and upload files part. 
Let me know if the errors are somewhat better and more descriptive now

mini-bomba:
I've picked my commits with utilities out of my local branch and pushed them onto main.
I've also renamed the `wrappedFetch` function to `fetchAndCheckStatus`, made a new `fetchJSON` variant and added those as protected functions on the `BaseScraperModule` class.
You should be able to use those without duplicating code after rebasing.

database/scrapers/buildings.ts
```diff
@@ -0,0 +1,202 @@
...
+          Readable.fromWeb(imageStream),
+          extension,
+        );
+      } catch (err) {
+        throw new Error(`failed to upload the file: ${imageKey}`, {
+          cause: err,
+        });
+      }
+    });
+  }
```
mini-bomba, line 163: (not an issue)
I'd usually wrap calls to the function in a semaphore.runTask() call, not the whole function definition
but I guess this works too? (time to make a semaphore decorator?)

database/scrapers/buildings.ts
```diff
@@ -0,0 +1,202 @@
...
+        }
+        return {
+          campusId: campus.id,
+          ...buildingEntry,
+        };
+      }),
+    );
+    task.update("Buildings created !");
+    //save changes for campuses
+    await Promise.all([...updatedCampuses].map((campus) => campus.save()));
```
mini-bomba, line 128: consider putting these .save() calls through the semaphore so that you don't accidentally upset adonis about the db pool being empty for too long...
(though on the other hand, adonis' internal db connection pool already functions like a semaphore)
either way this spead (`...`) operation here appears to be useless? you've only got one array and it should already have a .map() operartion defined

czaja307 (dismissed):
LGTM

mini-bomba (commented):
Literally the last minor issue, probably

simon-the-shark:
@ellipsis review this

app/models/building.ts
```diff
@@ -27,6 +27,8 @@ import Library from "./library.js";
   longitude: "number",
   haveFood: "boolean",
   cover: "string",
+  externalDigitalGuideMode: "string",
```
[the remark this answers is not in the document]
mini-bomba: możemy zbanować tego bota pliska?

database/scrapers/buildings.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  private async uploadCoverAndGetKey(
+    data: BuildingDraft,
+  ): Promise<string | undefined> {
+    const imageKey = data.cover;
+    if (imageKey === null) {
+      this.logger.warning(`no image for building: [${data.id}]`);
+      return;
+    }
```
mini-bomba, line 141: minor nitpick: Since the result of this function is going into the database, returning `null` will work as well and should be more explicit: this way you won't be able to forget a return statement and it will be clear that you're returning a `null` value.

mini-bomba (commented):
...and a few more issues/nitpicks

mini-bomba (approved):
:shipit: :fire:

mini-bomba (approved):

PR #106, opened 2025-02-13
feat: create scraper framework

mini-bomba:
This commit creates a single command that allows running multiple scrapers in series.
Scraper modules are defined in /database/scrapers.

some screenshots from testing:
two test "scrapers" completing successfully
image
one "scraper" throwing an error
image

closes #105

PR #112, opened 2025-02-22
feat: add command and scraper script for academic calendar

mateusz-plaska:
> [!IMPORTANT]
> Adds `AcademicCalendarScraper` to fetch, process, and store academic calendar and day swap data from external sources.
> 
>   - **New Feature**:
>     - Adds `AcademicCalendarScraper` class in `academic_calendar_scraper.ts` to fetch and process academic calendar and day swap data.
>     - Fetches data from `https://admin.topwr.solvro.pl/items/AcademicCalendarData` and `https://admin.topwr.solvro.pl/items/WeekExceptions`.
>     - Handles HTTP response errors for both data sources.
>   - **Data Processing**:
>     - Converts date strings to `DateTime` objects using `luxon`.
>     - Maps weekdays from string to `Weekday` enum.
>     - Filters and processes valid day swaps before storing.
>   - **Database Interaction**:
>     - Creates `AcademicCalendar` entries with fetched data.
>     - Associates day swaps with the academic calendar using `related("daySwaps").create()`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=Solvro%2Fbackend-topwr&utm_source=github&utm_medium=referral)<sup> for 06213fa653cbc48f0c2f54b0f359bfd61a3ab06a. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/database/scrapers/academic_calendar_scraper.ts
+++ b/database/scrapers/academic_calendar_scraper.ts
@@ -1,0 +1,89 @@
+import { DateTime } from "luxon";
+
+import { BaseScraperModule, TaskHandle } from "#commands/db_scrape";
+import { Weekday } from "#enums/weekday";
+import AcademicCalendar from "#models/academic_calendar";
+
+interface AcademicCalendarData {
+  data: {
+    id: number;
+    user_updated: string;
+    date_updated: DateTime;
+    semesterStartDate: DateTime;
+    examSessionStartDate: DateTime;
+    isFirstWeekEven: boolean;
+    examSessionLastDay: DateTime;
+  };
+}
+
+interface WeekExceptions {
+  data: {
+    id: number;
+    day: DateTime;
+    changedWeekday: string;
+    changedDayIsEven: boolean;
+  }[];
+}
+
+export default class AcademicCalendarScraper extends BaseScraperModule {
+  static name = "academic calendar scraper";
+  static description = "Import data about academic calendars and day swaps";
+
+  async run(task: TaskHandle): Promise<void> {
+    task.update("Fetching academic calendar and day swaps data");
+
+    const [academicCalendarResponse, daySwapsResponse] = await Promise.all([
+      fetch("https://admin.topwr.solvro.pl/items/AcademicCalendarData"),
+      fetch("https://admin.topwr.solvro.pl/items/WeekExceptions"),
+    ]);
+
+    if (!academicCalendarResponse.ok) {
+      throw new Error(
+        `Fetching academic calendar has failed. Status code: ${academicCalendarResponse.status}`,
+      );
+    }
+    if (!daySwapsResponse.ok) {
+      throw new Error(
+        `Fetching day swaps has failed. Status code: ${daySwapsResponse.status}`,
+      );
+    }
+
+    const [academicCalendarResult, daySwapsResult] = await Promise.all([
+      academicCalendarResponse.json() as Promise<AcademicCalendarData>,
+      daySwapsResponse.json() as Promise<WeekExceptions>,
+    ]);
+
+    const academicCalendar = await AcademicCalendar.create({
+      name: "",
+      semesterStartDate: academicCalendarResult.data.semesterStartDate,
+      examSessionStartDate: academicCalendarResult.data.examSessionStartDate,
+      examSessionLastDate: academicCalendarResult.data.examSessionLastDay,
+      isFirstWeekEven: academicCalendarResult.data.isFirstWeekEven,
+      createdAt: academicCalendarResult.data.date_updated,
+      updatedAt: academicCalendarResult.data.date_updated,
+    });
+
+    const weekdayMap = {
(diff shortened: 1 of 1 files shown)


czaja307:
check out #106, @mini-bomba created framework for running all the scripts at once

database/scrapers/academic_calendar_scraper.ts
```diff
@@ -0,0 +1,89 @@
...
+    changedDayIsEven: boolean;
+  }[];
+}
+
+export default class AcademicCalendarScraper extends BaseScraperModule {
+  static name = "academic calendar scraper";
+  static description = "Import data about academic calendars and day swaps";
+
+  async run(task: TaskHandle): Promise<void> {
+    task.update("Fetching academic calendar and day swaps data");
```
mini-bomba, line 33: You can set the initial task status message by setting the `taskTitle` static prop.
Since this script is just a single step, I don't think you need to output any additional status updates.

database/scrapers/academic_calendar_scraper.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    const academicCalendar = await AcademicCalendar.create({
+      name: "",
+      semesterStartDate: academicCalendarResult.data.semesterStartDate,
+      examSessionStartDate: academicCalendarResult.data.examSessionStartDate,
+      examSessionLastDate: academicCalendarResult.data.examSessionLastDay,
+      isFirstWeekEven: academicCalendarResult.data.isFirstWeekEven,
+      createdAt: academicCalendarResult.data.date_updated,
+      updatedAt: academicCalendarResult.data.date_updated,
+    });
```
mini-bomba, line 64: Could we maybe set some name for the calendar that isn't empty?
Something as stupid as "Imported calendar" would probably be better than an empty string (or at least would make it clear that someone should probably change it asap)
If you want to be fancy then maybe you could detect which semester it is and generate a proper calendar name like "2024/2025 Zima".

database/scrapers/academic_calendar_scraper.ts
```diff
@@ -0,0 +1,89 @@
...
+        date: daySwap.day,
+        changedWeekday:
+          weekdayMap[daySwap.changedWeekday as keyof typeof weekdayMap],
+        changedDayIsEven: daySwap.changedDayIsEven,
+      });
+    }
+
+    task.update(
+      "The academic calendar and day swaps data have been successfully migrated to the database",
+    );
```
mini-bomba, line 87: There's a "Done" success message printed automatically when you return from this function without throwing an error - this task update is not needed.
If you really want a custom success message (which imo should be kept as short as possible and "Done" works perfectly well in this case) then you can return a string from this function instead.

database/scrapers/academic_calendar_scraper.ts
```diff
@@ -0,0 +1,89 @@
...
+
+interface AcademicCalendarData {
+  data: {
+    id: number;
+    user_updated: string;
+    date_updated: DateTime;
+    semesterStartDate: DateTime;
+    examSessionStartDate: DateTime;
+    isFirstWeekEven: boolean;
+    examSessionLastDay: DateTime;
```
mini-bomba, line 15: An actual issue: The directus API returns strings, not `DateTime` objects. (in fact you cannot automatically deserialize such objects from JSON)
Even though adonis' orm knows how to cope with strings being passed as DateTimes, you should define those fields as strings (because that's what they actually are) and parse them if needed.
You can use `DateTime.fromISO()` to deserialize these DateTimes.

database/scrapers/academic_calendar_scraper.ts
```diff
@@ -0,0 +1,89 @@
...
+  };
+}
+
+interface WeekExceptions {
+  data: {
+    id: number;
+    day: DateTime;
+    changedWeekday: string;
+    changedDayIsEven: boolean;
+  }[];
```
mini-bomba, line 25: same issue here as above: define those as strings and parse them

mini-bomba (changes requested):
Really nice work! :fire: 
I've found just one minor issue and a few nitpicks.

simon-the-shark:
@ellipsis review this

database/scrapers/academic_calendar_scraper.ts
```diff
@@ -0,0 +1,98 @@
...
+      );
+    }
+
+    const [academicCalendarResult, daySwapsResult] = await Promise.all([
+      academicCalendarResponse.json() as Promise<AcademicCalendarData>,
+      daySwapsResponse.json() as Promise<WeekExceptions>,
+    ]);
+
+    const academicCalendar = await AcademicCalendar.create({
+      name: "2024/2025 Lato",
```
mini-bomba, line 56: Please use a name that is clearly a placeholder if you're not going to autodetect the proper calendar year and type.

mini-bomba (changes requested):

PR #113, opened 2025-02-22
feat(scrapers): contributors & changelogs schema scaper/migrator/whatever

### LINKED ISSUE #104: feat: contributors scraper

mini-bomba:
closes #104 

it is what is says it is
~~540~~ 501 lines of migration code
(i probably should've split it into separate functions)
~~i also stole some utils (from myself)~~
~~and updated node to 22.14, just in case~~
(i've pushed those two directly to main so that others can use them)


czaja307 (approved):
:fire: :shipit:

PR #119, opened 2025-03-04
feat(adminPanel): create reset password procedure

karolwolek:
In this pr I implemented:
* 2 routes for password reseting usefull in forgot password functionality for admin panel
* reset password command to operate from terminal
* setup mailer from adonis

In the end i gave up on doing anything frontend related if there is going to be a specific team for that (not that i didn't try, i even started learning react for that. BUT... doing any more complex changes to login page requires to write it from scratch and replace, so it's useless then to do it half way)
The logic behind forgot password functionality:
* you attach email to body in one request on the route, service checks for existence of any user with that email and tries to send an email (no feedback provided, only for lacking email) attaching under that user a hashed token in the database and defying its expiration date for 15min.
* The email contains url with token url-encoded. On request on another route with the same token url-encoded is sent new password. Service first checks for the validity of the token (and expiration date if first check passed) and the proceeds to change the password. Update for the same password will fail.

For some operations there are logs to console about actions.


app/models/user.ts
```diff
@@ -23,9 +24,36 @@ export default class User extends compose(BaseModel, AuthFinder) {
...
+      // deterministic hash for easier lookup
+      user.resetPasswordToken = crypto
+        .createHash("sha256")
+        .update(user.resetPasswordToken, "utf-8")
+        .digest("hex");
+    }
+  }
+
+  static compareTokens = scope((query, token: string) => {
+    void query.where(
```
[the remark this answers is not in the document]
karolwolek: szarap

app/services/reset_password_service.ts
```diff
@@ -0,0 +1,84 @@
...
+    if (
+      user.resetPasswordTokenExpiration !== null &&
+      user.resetPasswordTokenExpiration < DateTime.now()
+    ) {
+      user.resetPasswordToken = null;
+      user.resetPasswordTokenExpiration = null;
+      await user.save();
+      throw new UnathorizedException("Time out for request");
+    }
+    if (password === user.password) {
```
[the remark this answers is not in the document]
karolwolek: good point imo

app/services/reset_password_service.ts
```diff
@@ -0,0 +1,85 @@
...
+    if (
+      user.resetPasswordTokenExpiration !== null &&
+      user.resetPasswordTokenExpiration < DateTime.now()
+    ) {
+      user.resetPasswordToken = null;
+      user.resetPasswordTokenExpiration = null;
+      await user.save();
+      throw new UnathorizedException("Time out for request");
+    }
+    if (await hash.verify(user.password, password)) {
```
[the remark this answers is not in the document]
karolwolek: image
no.

czaja307 (changes requested):
good job, everything looks really solid - just some minor issues

app/services/reset_password_service.ts
```diff
@@ -0,0 +1,85 @@
...
+   *
+   * @param user user instance to associate token with
+   * @param expirationDate time till the token is valid. Pass DateTime instance using e.g. `DateTime.now().plus({minutes: 15})`
+   * @returns generated token (not hashed)
+   */
+  async generateResetPasswordToken(
+    user: User,
+    expirationDate: DateTime<true>,
+  ): Promise<string> {
+    const token = crypto.randomBytes(20).toString("hex");
```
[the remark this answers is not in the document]
karolwolek: using crypto.randomBytes with a 20-byte size is widely regarded as cryptographically secure and more than adequate for generating tokens, especially when you consider that these tokens have an expiration time. In a 15-minute window, the chance of generating a duplicate token is practically impossible

app/controllers/reset_passwords_controller.ts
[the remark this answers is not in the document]
karolwolek: If we choose more descriptive names for the methods, it could be an improvement. I was aiming to follow REST conventions, which is why I reserved the name user_controller for the actual operations on the model. But it can actually be better to do what you suggest

app/services/reset_password_service.ts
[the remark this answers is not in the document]
czaja307: thanks for clarification

start/routes.ts
```diff
@@ -37,11 +37,20 @@ const AcademicCalendarsController = () =>
...
   return { appName: env.get("APP_NAME"), version: env.get("APP_VERSION") };
 });
 
+router
+  .group(() => {
+    router.post("/", [ResetPasswordsController, "store"]);
+    router.put("/:token", [ResetPasswordsController, "update"]);
+  })
+  .prefix("admin/resetpassword"); //reset_password_service dependency
+
```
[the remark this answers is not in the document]
karolwolek: For other endpoints i think it's important to talk about that with mobile team because i think it's much more complicated topic. There is a thing about user system responsiveness and weirdos refreshing pages 100k times a minute. I will do it only for reset procedure for now

app/controllers/reset_passwords_controller.ts
```diff
@@ -0,0 +1,41 @@
...
+import BadRequestException from "#exceptions/bad_request_exception";
+import ResetPasswordService from "#services/reset_password_service";
+import { emailValidator } from "#validators/email";
+
+export default class ResetPasswordsController {
+  /**
+   * Page for password form submission
+   */
+  @inject()
+  async store(
```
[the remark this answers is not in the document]
mini-bomba: I'd assume `store` means `POST` without resource id, and that makes sense for a password reset endpoint (or any other endpoint that causes an action to be started/performed and not just some query)

app/services/reset_password_service.ts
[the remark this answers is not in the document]
mini-bomba: but you might also add a few more random bits and use UUID4, which is generally assumed to be safe from duplicates

.env.example
```diff
@@ -13,4 +13,9 @@ DRIVE_DISK=fs
 APP_URL=http://localhost:3333
 APP_NAME=ToPWR_Backend
 APP_VERSION=1.0.0
-SESSION_DRIVER=cookie
\ No newline at end of file
+SESSION_DRIVER=cookie
+SMTP_HOST=host.pl
+SMTP_PORT=outgoing_port
+SMTP_USERNAME=mail@mail.pl
```
mini-bomba, line 19: really minor: use `example.org` instead, that domain exists for this exact purpose

app/controllers/users_controller.ts
```diff
@@ -0,0 +1,40 @@
...
+  @inject()
+  async updatePassword(
+    { request, response }: HttpContext,
+    resetPasswordService: ResetPasswordService,
+  ) {
+    const token = request.param("token") as string;
+    const { password } = request.body() as { password: string | undefined };
+    if (password === undefined) {
+      throw new BadRequestException("No password provided");
+    }
```
mini-bomba, line 36: consider using validators to validate this request

app/mails/reset_password_notification.ts
```diff
@@ -0,0 +1,32 @@
...
+      <html>
+        <body style="font-family: sans-serif;">
+          <p>We received a request to reset your password for your account.</p>
+          <p>Click the link below to reset your password:</p>
+          <p><a href="${this.url}">Reset Password</a></p>
+          <p>If you did not request this, please ignore this email.</p>
+        </body>
+      </html>
+      `);
+  }
```
mini-bomba, line 31: improvement suggestion: does adonis support multipart emails? could we add a plaintext version?

database/migrations/1733076365295_create_users_table.ts
mini-bomba, line 1: minor suggestion: consider adding an index for the `reset_password_token` column
also use `text` instead of `string`

database/migrations/1741292347109_create_rate_limits_table.ts
```diff
@@ -0,0 +1,17 @@
+import { BaseSchema } from "@adonisjs/lucid/schema";
+
+export default class extends BaseSchema {
+  protected tableName = "rate_limits";
+
+  async up() {
+    this.schema.createTable(this.tableName, (table) => {
+      table.string("key", 255).notNullable().primary();
```
mini-bomba, line 8: use `text` instead of `string`, if it doesn't break things

start/routes.ts
```diff
@@ -37,11 +39,20 @@ const AcademicCalendarsController = () =>
...
   return { appName: env.get("APP_NAME"), version: env.get("APP_VERSION") };
 });
 
+router
+  .group(() => {
+    router.post("/", [ResetPasswordsController, "resetPassword"]);
+    router.put("/:token", [ResetPasswordsController, "updatePassword"]);
+  })
+  .use(resetPasswordThrottle)
+  .prefix("admin/resetpassword"); //reset_password_service dependency
```
mini-bomba, line 54: consider putting the `updatePassword` endpoint on a separate ratelimiter
we wouldn't want someone to get banned for clicking the reset link after finally getting the email on their 5th attempt
these two endpoints should be ratelimited for different reasons, so these should get their separate limiters: `resetPassword` to avoid spamming the email server, `updatePassword` to stop bruteforce attacks
ideally, only failed requests would count towards the `updatePassword` ratelimit.

mini-bomba (commented):
a bunch of minor issues/suggestions

database/migrations/1733076365295_create_users_table.ts
czaja307: isn't string sufficient in this case, as we know the length of the reset_password_token?

karolwolek: @czaja307 isn't string a varchar alias in lucid with postgres db which you highly discouraged ?

mini-bomba: yeah it is, that's my whole point
since there's (almost) no actual technical difference between text and string/varchar and string would be in our way should we decide to change the hashing function or the hash format in the future, we might as well use text for everything instead.

app/mails/reset_password_notification.ts
karolwolek: I am not familiar with that, and i will leave it like so, it's just a placeholder i guess for some fancier email - maybe frontend_for_backend team will have better ideas. For now it just works

database/migrations/1733076365295_create_users_table.ts
karolwolek: should all of those be changed to text as well ?
```ts
table.string("full_name").nullable();
      table.string("email", 254).notNullable().unique();
      table.string("password").notNullable();
```

czaja307: i guess so, i was just lazy and used the basic user model generated by adonis

start/routes.ts
karolwolek: I did as you suggested. for reset it is as it was previously but without blocking. For update i count only failed attempts

app/controllers/users_controller.ts
```diff
@@ -0,0 +1,67 @@
...
+    const {
+      params: { token },
+    } = result;
+
+    if (!token.isValid()) {
+      await resetPasswordService.destroyToken(token.user);
+      return response.unauthorized({
+        message: "Time out for request",
+      });
+    }
```
mini-bomba, line 60: "Time out for request" is a misleading message here, as it implies a server error.
Use something like "Invalid/expired token" (not saying which one exactly is also a good practice in these security-related endpoints)

Also tbh the validator should be automatically deleting expired tokens.

app/models/user.ts
```diff
@@ -23,9 +24,36 @@ export default class User extends compose(BaseModel, AuthFinder) {
...
+  static async hashToken(user: User) {
+    if (
+      user.$dirty.resetPasswordToken !== undefined &&
+      user.resetPasswordToken !== null
+    ) {
+      // deterministic hash for easier lookup
+      user.resetPasswordToken = crypto
+        .createHash("sha256")
+        .update(user.resetPasswordToken, "utf-8")
+        .digest("hex");
```
mini-bomba, line 49: we should have a helper function in utils to do a sha256 hash from a string, this is way too verbose.

app/services/reset_password_service.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+    if (await hash.verify(user.password, password)) {
+      throw new BadRequestException(
+        "Cannot change password for the already defined one",
+      );
+    }
+    user.password = password;
+    user.resetPasswordToken = null;
+    user.resetPasswordTokenExpiration = null;
+    await user.save();
```
mini-bomba, line 53: you've got a `destroyToken` function that does exactly this, use it

app/validators/reset_password_token.ts
```diff
@@ -0,0 +1,83 @@
...
+ * Will fail on invalid token formats OR no user assigned to the token.
+ * Expiration date must be handled by checking `isValid()` method on output
+ */
+export const resetPasswordTokenValidator = vine.compile(
+  vine.object({
+    params: vine.object({
+      token: vine.string().resetPasswordToken(),
+    }),
+  }),
+);
```
mini-bomba, line 83: as noted earlier: I think objects in `#validators/*` should only contain reusable components, define single-use validators where they'll be used

commands/reset_password.ts
```diff
@@ -0,0 +1,27 @@
...
+    "Tries to reset the password for user associated with given email address";
+
+  static options: CommandOptions = {
+    startApp: true,
+  };
+
+  @args.string({
+    argumentName: "email",
+    description:
+      "Email address associated with user for whom the resetting is done",
```
mini-bomba, line 19: this description is fire

database/migrations/1741868507062_alter_user_reset_token_and_string_to_text.ts
mini-bomba, line 1: this migration is missing a rollback function

mini-bomba (changes requested):
minor issues

app/controllers/users_controller.ts
karolwolek: I don't think it's a good idea, too many thinks on the back of one functionality concept. Validators are meant to check data only. I was trying to follow single dependency. Imagine someone reading the code snippet where the validator is used, how could he know that the validators is also doing some save/update action on the model itself.

mini-bomba: tru

app/validators/password.ts
```diff
@@ -0,0 +1,52 @@
...
+  vine.object({
+    password: vine
+      .string()
+      .minLength(8)
+      .use(lowerCaseLetterRule())
+      .use(upperCaseLetterRule())
+      .use(digitRule())
+      .bail(false),
+  }),
+);
```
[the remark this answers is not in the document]
karolwolek: Could't this validator be used in elsewhere, probably where the user new user would be created or anything else ? The rules were just a proposition but i think it's totally fine to leave it here as it can be used in other admin panel components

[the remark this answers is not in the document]
mini-bomba: what I mean is that the `passwordValidator` here should not assume that it's in an object with a field that will be named `password`, it rather should only define the rules for what a password should be, which will be used by locally defined validators for specific requests
it was fine to create whole object validators for scopes, because multiple controllers would always use the same parameter structure and the validator would only be used in that context.
here, the password validator will probably will be used in multiple controllers that all take in different input structures, and the field name to validate might also be different (think a password change endpoint, that would take fields called `old`/`newPassword` instead)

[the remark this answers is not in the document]
czaja307: https://docs.adonisjs.com/guides/basics/validation#creating-validators
image
i think that we don't have to worry too much about this, we can think about some refactoring when we will need it (but i'd say that copy-paste works fine as well)

app/services/reset_password_service.ts
```diff
@@ -0,0 +1,81 @@
...
+    ) {
+      user.resetPasswordToken = null;
+      user.resetPasswordTokenExpiration = null;
+      await user.save();
+    }
+    if (await hash.verify(user.password, password)) {
+      throw new BadRequestException(
+        "Cannot change password for the already defined one",
+      );
+    }
```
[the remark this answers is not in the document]
czaja307: at least then you know what your password is :p

app/models/user.ts
czaja307: it's seems fine to me and i don't really thing that a 3 line function is that necessary (but it may be just my preference)

app/validators/password.ts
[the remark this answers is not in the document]
karolwolek: I left password validator untached and added a VineString rules combined:
``` ts
export const passwordRules = vine
  .string()
  .minLength(8)
  .use(lowerCaseLetterRule())
  .use(upperCaseLetterRule())
  .use(digitRule())
  .bail(false);
  ```

File: app/validators/reset_password_token.ts (hunk 15 above)
Comment (karolwolek) [reply app/validators/reset_password_token.ts:83]: Following @czaja307 comment on the reusable validators i don't think it's bad idea to seperate it and leave it in here. As it declutters the controller file and clearly indicates the usage context. If you do not agree with this guys let me know

File: app/models/user.ts [hunk 19]
```diff
@@ -23,9 +26,40 @@ export default class User extends compose(BaseModel, AuthFinder) {
...
+    ) {
+      // deterministic hash for easier lookup
+      user.resetPasswordToken = sha256(user.resetPasswordToken);
+    }
+  }
+
+  static compareTokens = scope((query, token: string) => {
+    void query.where(
+      "reset_password_token",
+      crypto.createHash("sha256").update(token, "utf-8").digest("hex"),
```
mini-bomba, line 55: minor: you can use the sha256 helper function here

app/services/reset_password_service.ts
```diff
@@ -0,0 +1,64 @@
...
+      return;
+    }
+    const token = await this.generateResetPasswordToken(
+      user,
+      DateTime.now().plus({ minutes: 15 }),
+    );
+    await mail.send(
+      new ResetPasswordNotification(
+        email,
+        `${env.get("APP_URL")}/admin/resetpassword/${token}`,
```
mini-bomba, line 28: this URL does not appear to be valid
image

mini-bomba (changes requested):
apart from that 1 minor and 1 less minor issue:
optional side quest: rebase on top of main and replace all uses of `response.*` with throwing an appropriate error from `#exceptions/http_exceptions` (create more classes if needed)

app/models/user.ts
karolwolek: forgot about that

app/services/reset_password_service.ts
karolwolek: it's a put request not get

mini-bomba: but you're sending it in an email, and browsers use GET for opening links

karolwolek: tru

karolwolek: i didn't think it through then. Should there be a get route aswell (possible with the same structure) that renders some page and on submit it's using this url as a put request ?

mini-bomba: that's one solution.
alternatively, check if adminjs can't perhaps provide you with a frontend for the user to finish the password reset.

app/exceptions/http_exceptions.ts
```diff
@@ -39,3 +39,23 @@ export class NotImplementedException extends BaseError {
     });
   }
 }
+
+export class TooManyRequestsException extends BaseError {
+  constructor(message?: string, options?: BaseErrorOptions) {
+    super(message ?? "To many requests", {
+      code: "E_TO_MANY_REQUESTS",
```
mini-bomba, line 46: <details>
<summary>minor spelling mistake</summary>

minor spelling mistake
</details>

mini-bomba (approved):
:fire: :shipit:
time to test in prod at 2 am
what could go wrong

PR #121, opened 2025-03-07
Feat/admin panel upload photos

Leadman5555:
Let me know if this meets the expectations.
Main changes:
- all models with 'cover' attribute have now the function to upload a photo as their cover via dropping or file browsing. All that needs to be done to enable it for a model is to pass a single boolean value as true in resourceBuilder
- information about the cover is saved in db, while the photo itself on disk; Both are updated/deleted as needed
- fileService was made static and refactored, alongside with adding a model for Files
- custom component for uploading photos (at least I didn't have to do the css). We cannot use the built-in uploadFeature since we want to use our fileService for uploads. 
image

- refactored the admin panel heavily, mainly to reduce the awful boiler plate code for resources and make adding new ones far easier and less error-prone
- I hate that ESLint plugin, could we please relax some of the requirements? Especially braces for if one-liners, detailed boolean conditions and any type rules...


mini-bomba (changes requested):
I don't even know how to begin to review this PR.
You appear to have made a lot of vaguely related changes all in the same PR, and even in the same commits.
Please split these changes into separate PRs - don't do refactors in a "feature" PR/commit.
Use `npm ci` instead of `npm i` to avoid updating the lockfile. If you want to update dependencies, please do so in a separate PR, while ensuring that nothing broke as a result.

Leadman5555:
Fine, I'll split it into 3 smaller ones, although they need to be accepted one by one. One for file service, one for dropbox component and one for admin panel refactor

Leadman5555:
First one is up

PR #123, opened 2025-03-10
refactor(fileService): save info about stored files in the database

Leadman5555:
Part one out of three from feat/adminPanel-uploadPhotos

----

> [!IMPORTANT]
> Refactor `FilesService` to store file metadata in the database and update related components to use new service methods.
> 
>   - **Behavior**:
>     - `FilesService` methods now store file metadata in the database using the `FileEntry` model.
>     - `uploadMultipartFile`, `uploadStream`, `uploadFromMemory`, `uploadLocalFile`, and `replaceWithMultipartFile` methods in `files_service.ts` now use transactions to save file metadata.
>     - `getFileUrl` and `deleteFileWithKey` methods in `files_service.ts` handle file metadata retrieval and deletion.
>   - **Models**:
>     - New `FileEntry` model in `file_entry.ts` to store file metadata with methods `createNew` and `fetchKeyWithExtension`.
>   - **Controllers**:
>     - `FilesController` in `files_controller.ts` updated to use static methods of `FilesService`.
>   - **Exceptions**:
>     - New exceptions `FileServiceFilePersistError`, `FileServiceFileDiskDeleteError`, and `FileServiceFileMetadataDeleteError` in `file_service_errors.ts`.
>   - **Database**:
>     - New migration `1741334906365_create_file_entries_table.ts` to create `file_entries` table.
>   - **Misc**:
>     - Removed `@inject()` decorators in `files_controller.ts`.
>     - Updated `buildings.ts`, `contributors.ts`, and `about_us_seeder.ts` to use new `FilesService` methods.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=Solvro%2Fbackend-topwr&utm_source=github&utm_medium=referral)<sup> for 2b7349b5232420885b141205638ae0e16a214191. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/app/models/file_entry.ts
+++ b/app/models/file_entry.ts
@@ -17,11 +17,11 @@ export default class FileEntry extends BaseModel {
   @column.dateTime({ autoCreate: true, autoUpdate: true })
   declare updatedAt: DateTime;
 
-  get diskKey() {
+  get keyWithExtension() {
     return `${this.id}.${this.fileExtension}`;
   }
 
-  public static createFileEntry(extname: string | undefined): FileEntry {
+  public static createNew(extname: string | undefined): FileEntry {
     const fileEntry = new FileEntry();
     fileEntry.id = randomUUID();
     fileEntry.fileExtension =
@@ -29,29 +29,15 @@ export default class FileEntry extends BaseModel {
     return fileEntry;
   }
 
-  public static async deleteByKeyAndReturnResult(
+  public static async fetchKeyWithExtension(
     key: string,
-  ): Promise<boolean> {
-    return FileEntry.query()
-      .where("id", key)
-      .delete()
-      .then((res: number[]) => res.length > 0 && res[0] > 0);
-  }
-
-  public static async updateFileType(
-    newFileType: string,
-    existingFileKey: string,
-  ) {
-    return FileEntry.query()
-      .where("id", existingFileKey)
-      .update({ file_type: newFileType });
-  }
-
-  public static async getExtensionIfExists(key: string) {
+  ): Promise<string | null> {
     return FileEntry.query()
       .select("file_extension")
       .where("id", key)
       .first()
-      .then((value) => value?.fileExtension);
+      .then((value) =>
+        value !== null ? `${key}.${value.fileExtension}` : null,
+      );
   }
 }
--- a/app/models/file_entry.ts
+++ b/app/models/file_entry.ts
@@ -1,4 +1,5 @@
 import { DateTime } from "luxon";
+import { randomUUID } from "node:crypto";
 import type { UUID } from "node:crypto";
 
 import { BaseModel, column } from "@adonisjs/lucid/orm";
@@ -7,8 +8,8 @@ export default class FileEntry extends BaseModel {
   @column({ isPrimary: true })
   declare id: UUID;
 
-  @column({ columnName: "file_type" })
-  declare fileType: string;
+  @column({ columnName: "file_extension" })
+  declare fileExtension: string;
 
   @column.dateTime({ autoCreate: true })
   declare createdAt: DateTime;
@@ -16,8 +17,16 @@ export default class FileEntry extends BaseModel {
   @column.dateTime({ autoCreate: true, autoUpdate: true })
   declare updatedAt: DateTime;
 
-  public static async deleteByKey(key: string) {
-    return FileEntry.query().where("id", key).delete();
+  get diskKey() {
+    return `${this.id}.${this.fileExtension}`;
+  }
+
+  public static createFileEntry(extname: string | undefined): FileEntry {
+    const fileEntry = new FileEntry();
+    fileEntry.id = randomUUID();
+    fileEntry.fileExtension =
+      extname !== undefined && extname.length > 0 ? extname : "bin";
+    return fileEntry;
   }
 
   public static async deleteByKeyAndReturnResult(
@@ -38,9 +47,11 @@ export default class FileEntry extends BaseModel {
       .update({ file_type: newFileType });
   }
 
-  public static async getExtensionIfExists(
-    key: string,
-  ): Promise<string | undefined> {
-    return FileEntry.find(key).then((res) => res?.fileType);
+  public static async getExtensionIfExists(key: string) {
+    return FileEntry.query()
+      .select("file_extension")
+      .where("id", key)
+      .first()
+      .then((value) => value?.fileExtension);
   }
 }
--- a/app/services/file_entry.ts
+++ b/app/services/file_entry.ts
@@ -1,0 +1,46 @@
+import { DateTime } from "luxon";
+import type { UUID } from "node:crypto";
+
+import { BaseModel, column } from "@adonisjs/lucid/orm";
+
+export default class FileEntry extends BaseModel {
+  @column({ isPrimary: true })
+  declare id: UUID;
+
+  @column({ columnName: "file_type" })
+  declare fileType: string;
+
+  @column.dateTime({ autoCreate: true })
+  declare createdAt: DateTime;
+
+  @column.dateTime({ autoCreate: true, autoUpdate: true })
+  declare updatedAt: DateTime;
+
+  public static async deleteByKey(key: string) {
+    return FileEntry.query().where("id", key).delete();
+  }
+
+  public static async deleteByKeyAndReturnResult(
+    key: string,
+  ): Promise<boolean> {
+    return FileEntry.query()
+      .where("id", key)
+      .delete()
+      .then((res: number[]) => res.length > 0 && res[0] > 0);
+  }
+
+  public static async updateFileType(
+    newFileType: string,
+    existingFileKey: string,
+  ) {
+    return FileEntry.query()
+      .where("id", existingFileKey)
+      .update({ file_type: newFileType });
+  }
+
+  public static async getExtensionIfExists(
+    key: string,
+  ): Promise<string | undefined> {
+    return FileEntry.find(key).then((res) => res?.fileType);
+  }
+}
--- a/app/services/files_service.ts
+++ b/app/services/files_service.ts
@@ -3,9 +3,11 @@ import { Readable } from "node:stream";
 
 import { MultipartFile } from "@adonisjs/core/types/bodyparser";
 import drive from "@adonisjs/drive/services/main";
+import db from "@adonisjs/lucid/services/db";
 
 import {
-  FileServiceFileDeleteError,
+  FileServiceFileDiskDeleteError,
+  FileServiceFileMetadataDeleteError,
   FileServiceFilePersistError,
   FileServiceFileReadError,
   FileServiceFileUploadError,
@@ -17,10 +19,6 @@ export default class FilesService {
     return extname !== undefined && extname.length > 0 ? extname : "bin";
   }
 
-  private static mapToDiskKey(fileKey: string, fileType: string): string {
-    return `${fileKey}.${fileType}`;
-  }
-
   /**
    * Uploads a multipart file to storage.
    *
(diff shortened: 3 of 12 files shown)


app/services/files_service.ts
```diff
@@ -10,13 +10,41 @@ import {
   FileServiceFileReadError,
   FileServiceFileUploadError,
 } from "#exceptions/file_service_errors";
+import FileEntry from "#services/file_entry";
+
+interface FileInformation {
+  fileInstance: FileEntry;
+  diskKey: string;
+}
```
mini-bomba, line 18: maybe make `diskKey` a getter on the `FileEntry` model instead?

app/services/files_service.ts
```diff
@@ -10,13 +10,41 @@ import {
...
+  private static getExtension(extname: string | undefined): string {
+    return extname !== undefined && extname.length > 0 ? extname : "bin";
+  }
+
+  private static mapToFileInstance(extname: string | undefined): FileEntry {
+    const fileEntry = new FileEntry();
+    fileEntry.id = randomUUID();
+    fileEntry.fileType = this.getExtension(extname);
+    return fileEntry;
+  }
```
mini-bomba, line 30: `mapToFileInstance` seems like the wrong name for this function, as it allocates a new file entry instead of mapping an existing one to a model instance.
also maybe this should be a factory method on `FileEntry` instead?

app/services/files_service.ts
```diff
@@ -27,16 +55,55 @@ export default class FilesService {
...
+   *
+   * Use this if you're handling an API request and received a `MultipartFile` from adonis
+   * @param file - The file to upload
+   * @param existingFileKey - the key of file you wish to replace, without the extension name
+   * @throws {FileServiceFileUploadError} There was an issue uploading the file. Check the cause prop for details.
+   */
+  static async replaceWithMultipartFile(
+    file: MultipartFile,
+    existingFileKey: string,
+  ): Promise<string> {
```
mini-bomba, line 80: what's the use case for replacing existing files?
imo uploaded files should be immutable and replacing files should involve uploading a new one, replacing existing references and finally deleting the original file.
that approach would also make caching trivial.

app/services/files_service.ts
```diff
@@ -27,16 +55,55 @@ export default class FilesService {
...
+    const fileInfo = FilesService.createFileInformation(file.extname);
     try {
-      await file.moveToDisk(key);
-      return key;
+      await fileInfo.fileInstance.save();
+      await file.moveToDisk(fileInfo.diskKey);
+      return fileInfo.diskKey;
     } catch (error) {
       throw new FileServiceFileUploadError(error as Error);
     }
```
mini-bomba, line 66: `FileServiceFileUploadError` should only wrap errors related to issues uploading the file to the intended storage space, as its error message mentions.
But `.save()` on a model might fail due to a database issue, which is a different issue and as such should get a different context message and a separate error class.
Please put the database operations in a separate try/catch block and assign a separate error class for these errors.

app/services/files_service.ts
```diff
@@ -127,15 +199,47 @@ export default class FilesService {
...
+      const extension = await FileEntry.getExtensionIfExists(key);
+      if (extension !== undefined) {
+        await FileEntry.deleteByKey(key);
+        await drive.use().delete(this.mapToDiskKey(key, extension));
+        return true;
+      }
+      return false;
     } catch (error) {
       throw new FileServiceFileDeleteError(error as Error);
     }
```
mini-bomba, line 239: since the primary key only covers the `key` column, there should be no difference between fetching/deleting by a bare key and a full key.
also what about a function to delete using a `FileEntry` instance?

database/migrations/1741334906365_create_file_entries_table.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export default class extends BaseSchema {
+  protected tableName = "file_entries";
+
+  async up() {
+    this.schema.createTable(this.tableName, (table) => {
+      table.uuid("id").primary();
+      table.string("file_type").notNullable();
+      table.timestamp("created_at").notNullable();
+      table.timestamp("updated_at").nullable();
```
mini-bomba, line 11: updated_at should not be nullable

database/migrations/1741334906365_create_file_entries_table.ts
```diff
@@ -0,0 +1,18 @@
+import { BaseSchema } from "@adonisjs/lucid/schema";
+
+export default class extends BaseSchema {
+  protected tableName = "file_entries";
+
+  async up() {
+    this.schema.createTable(this.tableName, (table) => {
+      table.uuid("id").primary();
+      table.string("file_type").notNullable();
```
mini-bomba, line 9: use text instead of string

mini-bomba (changes requested):
as always, a bunch of minor issues and minor-er issues

major highlights:
- error handling needs to be well thought out - don't mix different error types under the same context message/error class
- attach methods/getters related to some object to that object
- avoid creating static methods that run queries and discard the fetched model instance afterwards - the caller might already have or might want to use an instance of that model

app/services/file_entry.ts
```diff
@@ -0,0 +1,46 @@
...
+import type { UUID } from "node:crypto";
+
+import { BaseModel, column } from "@adonisjs/lucid/orm";
+
+export default class FileEntry extends BaseModel {
+  @column({ isPrimary: true })
+  declare id: UUID;
+
+  @column({ columnName: "file_type" })
+  declare fileType: string;
```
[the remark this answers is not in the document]
Leadman5555: File extension does indeed fit better. As for the column name, it's safer to add it in case its mapped to "fileType" instead. It doesn't really complicated anything anyway

app/services/file_entry.ts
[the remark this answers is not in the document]
Leadman5555: My reasoning was that since fileEntry was only used in one place - FileService - it was better to put it there instead of the shared models directory. But that's your call, I moved it to models

app/services/files_service.ts
Leadman5555: Considered it done

database/migrations/1741334906365_create_file_entries_table.ts
Leadman5555: I disagree with this. We are persisting file extensions, which are short strings, not above 255 chars - limit of file name length, so why should we use a text type column here?

app/services/files_service.ts
Leadman5555: Such approach would indeed make things simpler. But not without a tradeoff. We would always need to delete a record and persist a new one, which isn't really needed if we just replace the underlying photo. At other times, we only need to change the file extension in the database, without the need to perform more modifying queries.

app/services/file_entry.ts
```diff
@@ -0,0 +1,46 @@
...
+    return FileEntry.query()
+      .where("id", existingFileKey)
+      .update({ file_type: newFileType });
+  }
+
+  public static async getExtensionIfExists(
+    key: string,
+  ): Promise<string | undefined> {
+    return FileEntry.find(key).then((res) => res?.fileType);
+  }
```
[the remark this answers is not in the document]
Leadman5555: The getExtensionIfExists() was changed in new commit. As for other methods - yes, you could operate on an existing instance. An existing one. But in cases where those methods are used, there is no existing instance. Which means we would need to first fetch one and then update it. But in my approach, we don't need to fetch at all - we can just update. It's more efficient that way as we make one query less.

Leadman5555:
New version is out. Most concerns were addressed. As for the remaining few, I'd love to hear your on my comments

database/migrations/1741334906365_create_file_entries_table.ts
mini-bomba: this was a topic of a previous discussion, but the TL:DR is: we're using postgres and there is no difference in the underlying database storage type between `string` and `text` and changing the max length is a pain, should we ever want to do that. (at least in older psql versions)

app/services/files_service.ts
mini-bomba: yeah but then we wouldn't be able to tell the client to cache assets effectively forever - it would increase potential server load and make the clients have to send additional requests to validate their caches
as for file extension changes, your database model already defined the uuid as a primary key and the extension as a separate field, so it would be safe to just change the file extension in the database - i don't mind that part becoming mutable metadata
we could even make the storage controller ignore the file extension (and have it set the extension in the downloaded filename via headers), we could define foreign key constraints on file key fields in existing models, we could remove the extension from local filenames entirely

app/models/file_entry.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  public static createFileEntry(extname: string | undefined): FileEntry {
+    const fileEntry = new FileEntry();
+    fileEntry.id = randomUUID();
+    fileEntry.fileExtension =
+      extname !== undefined && extname.length > 0 ? extname : "bin";
+    return fileEntry;
+  }
+
+  public static async deleteByKeyAndReturnResult(
```
mini-bomba, line 32: since that other delete variant no longer exists (and very well, cause you should always be checking if the entry actually got deleted) we can give this function a saner name, such as `deleteByKey`

mini-bomba (changes requested):
some more minor issues
with some minor-er code style and naming nitpicks

database/migrations/1741334906365_create_file_entries_table.ts
mini-bomba, line 9: oh you also forgot to update the column name in the migration script

Leadman5555: image
In the newer versions, I don't know about the older ones, you can achieve that in one query, at least when changing from shorter `varchar` to nearly unlimited `text` - it's probably harder the other way around. In my opinion leaving it as `varchar` shows that we know that file extension cannot be 60K characters long. But I'll change it to text to keep the convention from other tables.

mini-bomba: the issue was not the amount of queries, but rather the type of locks the query required and the cost of that query

app/services/files_service.ts
```diff
@@ -127,15 +198,47 @@ export default class FilesService {
...
+      );
+      if (recordExists) {
+        await drive.use().delete(diskKey);
+        return true;
+      }
+      return false;
     } catch (error) {
       throw new FileServiceFileDeleteError(error as Error);
     }
   }
```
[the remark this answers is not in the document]
Leadman5555: It was used during a entity delete hook in the 3rd part of this PR. Essentially it avoided making unneeded calls to the DB to fetch the file_extension. So essentially unless someone directly modified the to be deleted entity field in the DB, it would never fail. Although someone might misuse it later on, so maybe you are right, the extra overhead might be worth it.

app/services/files_service.ts
```diff
@@ -27,16 +29,66 @@ export default class FilesService {
...
     try {
-      await file.moveToDisk(key);
-      return key;
+      const diskKey = fileEntry.diskKey;
+      await file.moveToDisk(diskKey);
+      return diskKey;
     } catch (error) {
       throw new FileServiceFileUploadError(error as Error);
     }
   }
```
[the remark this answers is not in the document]
Leadman5555: My bad, I assumed it's automatically wrapped in a transaction context (as most things seem to be doing something by themselves in adonis)

app/services/files_service.ts
Leadman5555: The last approach seems to be an interesting one. It would also mean that any file information *needs* to be saved in the database, otherwise you'd need to guess the extension. I'll try to introduce this approach

app/services/files_service.ts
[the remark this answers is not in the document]
mini-bomba: wrapping the function in a transaction context would be even better, didn't think about that

app/models/file_entry.ts
```diff
@@ -38,9 +47,11 @@ export default class FileEntry extends BaseModel {
...
-    key: string,
-  ): Promise<string | undefined> {
-    return FileEntry.find(key).then((res) => res?.fileType);
+  public static async getExtensionIfExists(key: string) {
+    return FileEntry.query()
+      .select("file_extension")
+      .where("id", key)
+      .first()
+      .then((value) => value?.fileExtension);
   }
```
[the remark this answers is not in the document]
mini-bomba: I just think it's bad practice to do partial selects of objects when we're using an ORM.
We have all the ORM models, so let's actually use them.
And the database will have to find the whole row anyway, the only possible difference I can think of would be a minor serialization/network/deserialization overhead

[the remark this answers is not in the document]
Leadman5555: ORM is good, but forced ORM like in Lucid that always wraps the result is quite annoying, at least for me. A shame there are no entity views in Lucid, at least from what I seem to have found...

app/services/files_service.ts
Leadman5555: After some trial and error, it's not a good approach. When later requesting files from server, the url would need to be without extension and the file type would be unknown

database/migrations/1741334906365_create_file_entries_table.ts
mini-bomba, line 1: can you import existing files into the database in this migration?

czaja307: @mini-bomba what do you think about just re-running migration script it will be much easier and I doubt that there are any serious users by now

mini-bomba: i guess

mini-bomba (approved):
:shipit:

PR #129, opened 2025-03-16
feat(AdminPanel): add photo_dropbox component

Leadman5555:
# Part 2 out of 3 for feat(AdminPanel): upload photos
Currently the component is not visible anywhere - it will be linked to models in part 3 of this pull request series.
It will look exactly like this, displaying the preview of selected image:
image

### DIFF
--- a/app/admin/components/photo_dropbox.tsx
+++ b/app/admin/components/photo_dropbox.tsx
@@ -1,0 +1,52 @@
+import { Box, DropZone, DropZoneItem, Label } from "@adminjs/design-system";
+import { BasePropertyProps } from "adminjs";
+import { File } from "node:buffer";
+import React, { FC } from "react";
+
+async function fileToBase64(file: File): Promise<string> {
+  const buffer = await file.arrayBuffer();
+  return Buffer.from(buffer).toString("base64");
+}
+
+//Maximum file size allowed in dropbox in MB
+const MAX_FILE_SIZE = 1024 * 1024 * 5;
+//MIME types that the dropbox will allow to upload
+const ALLOWED_FILE_TYPES = ["image/png", "image/jpeg", "image/webp"];
+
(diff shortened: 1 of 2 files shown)


app/admin/components/photo_dropbox.tsx
```diff
@@ -0,0 +1,52 @@
...
+    if (files.length > 0) {
+      void fileToBase64(files[0]).then((bs) =>
+        setPreview(`data:image/jpeg;base64,${bs}`),
+      );
+      onChange(property.name, files[0]);
+    } else {
+      setPreview(undefined);
+      onChange(property.name, null);
+    }
+  };
```
mini-bomba, line 32: minor suggestion: I think [creating a blob: link](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL_static) from a file should be more efficient than base64 encoding the contents, at the price of having to properly free the file after it's no longer used. 
The URL allocation method not being async would likely be a bonus here.
... but you'd need to either use the other type of component declaration, or use some hook to run code on unmount.

mini-bomba (commented):
:fire:
i've got just one smol suggestion to consider

app/admin/components/photo_dropbox.tsx
Leadman5555: I have experimented a bit with the DropBox element and it appears that adminJs displays the preview by itself, AS LONG AS, you don't add a dropZoneItem element.
If you add it and set it to display a src, it will replace the default one. On the other hand, if you set it to display a file, there will be two duplicate previews displayed.
I love that there isn't a single word about it in the docs...
Either way, this should solve the issue on our side, although I have no idea how exactly adminJS generates the preview on it's own

mini-bomba (approved):
:fire: :shipit:

PR #130, opened 2025-03-16
refactor(admin panel): uploading photos and resource creation

Leadman5555:
# Part 3 out of 3 for issue: AdminPanel uploadPhotos


app/admin/resources/resource_factory.ts
```diff
@@ -0,0 +1,252 @@
...
+  }
+
+  private static replaceExistingCover(
+    currentCover: string,
+    newCover: MultipartFile,
+  ): Promise<string> {
+    return FilesService.deleteFileWithKey(currentCover).then(() => {
+      return FilesService.uploadMultipartFile(newCover);
+    });
+  }
```
mini-bomba, line 251: one of my linters (rightfully) suggests that this could be converted to an async function
also you probably should first upload then delete, so that you don't end up with no file if the upload fails or takes a long time

app/admin/resources/resource_factory.ts
```diff
@@ -0,0 +1,252 @@
...
+
+  public registerResource(resourceBuilder: ResourceBuilder) {
+    this.registeredResources.push(resourceBuilder);
+  }
+
+  public buildResources() {
+    return this.registeredResources
+      .map((resourceBuilder) => this.createResources(resourceBuilder))
+      .flat(1);
+  }
```
mini-bomba, line 101: add an explicit return type here

app/admin/resources/resource_factory.ts
```diff
@@ -0,0 +1,252 @@
...
+
+  private static createResource(
+    resourceModel: LucidModel,
+    resourceNavigation: ResourceNavigation,
+    additionalProperties?: Record<string, PropertyOptions>,
+    additionalActions?: ActionMap,
+    additionalOptions?: ResourceOptions,
+    addImageHandling = false,
+  ): ResourceWithOptions {
```
mini-bomba, line 125: with so many parameters (some optional) I'd consider passing everything in an object to immitate keyword arguments and make it easier to read what's going on at the call site (plus you might be able to use `...` to merge both objects)

app/admin/resources/resourceRegisters/about_us.ts
mini-bomba, line 1: applies to all resourceRegisters:
- you probably don't need a factory function for these builder objects, a constant value would probably work (unless i'm missing something)
- use snake_case in file/directory names

app/admin/resources/resource_factory.ts
```diff
@@ -0,0 +1,252 @@
...
+  new: false,
+  list: true,
+};
+
+const hideOnShow = {
+  list: false,
+  show: false,
+  new: true,
+  edit: true,
+};
```
mini-bomba, line 61: very minor: use the same property order in these two objects

mini-bomba (commented):
minor
issues

app/admin/resources/resource_factory.ts
mini-bomba: actually thinking about this further: you probably should upload in the before hook and delete in the after hook

app/admin/resources/resource_factory.ts
```diff
@@ -0,0 +1,254 @@
...
+        _: ImprovedRecordActionResponse,
+        __: ActionRequest,
+        context: ContextWithCover,
+      ): Promise<RecordActionResponse> => {
+        const oldCover: string | undefined = context.old_cover;
+        if (typeof oldCover === "string") {
+          await FilesService.deleteFileWithKey(oldCover);
+        }
+        return _;
+      },
```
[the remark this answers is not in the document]
mini-bomba: also log something if the file didn't get deleted (aka the delete function returned false)
if we can log something here that is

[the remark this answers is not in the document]
mini-bomba: that's exactly what I'm asking for. it's a rare case, but I think it's worth logging just in case - silent errors aren't fun to debug
anyway, do we have access to adminjs' or adonis' loggers here? if not, you should be able to use `console.warn`, which shouldn't trigger our linter.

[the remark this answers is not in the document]
mini-bomba: I assumed that these hook functions were running on the server (as otherwise how would it have access to the server's file system?)
since this is a server-side issue, I think the logs should be emitted on the server side.
server side logs are also much more reliably persisted - you can browse the container logs and they're saved for at least the lifetime of the container, while the client side logs get cleared every time the client navigates to a different page (which often is a shorter time than the server container lifetime)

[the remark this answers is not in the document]
mini-bomba: if these run on the client, then the upload code has no right to work.
if these run on the server, then it doesn't matter whether you use `console` or adonis' logger, it's going to the server's console

app/admin/resources/resource_factory.ts
```diff
@@ -0,0 +1,260 @@
...
+        _: ActionRequest,
+        context: ContextWithCover,
+      ): Promise<RecordActionResponse> => {
+        const oldCover: string | undefined = context.old_cover;
+        if (typeof oldCover === "string") {
+          const result = await FilesService.deleteFileWithKey(oldCover);
+          if (!result) {
+            logger.info(
+              "Cover edit: Old file does not exist - no file was deleted",
+            );
```
mini-bomba, line 236: the `warning` log level is more fitting for this kind of message

mini-bomba (changes requested):
uhm yeah: I decided to actually test the admin panel on this branch
it's broken.
- you assume that every image key field will always be called `cover`. simply not true
  - the contributor table uses `photoKey`, screenshots use `imageKey`, guide articles use `imagePath`
  - yes, these are inconsistencies in our naming scheme that are the result of every schema being made by a different person. it's not good, but your code should at least be flexible enough to support differently named fields and possibly multiple of them in one object.
- your image upload component does not show the current image and there's no way to delete it if desired.
- a few minor ones:
  - your upload field is always called `uploadPhoto`. the name should probably be customizable and the default should be a bit more "human" (or make it required and set no defaults)
  - the upload field is always the first item in the edit menu
  - there's no margin between the upload widget and the item below
    margin missing below upload widget

mini-bomba:
yes, everything that references a file or image is stored in our files service.
I guess the naming mess was due to us not knowing how we'll actually store files back when we were designing the schemas. (I even wanted a centralised file metadata table back then, to link fields via foreign keys to)

renaming the fields might be possible but would need to be consulted (cc @czaja307 )

whether we decide to rename the fields or not, your code should support differently named image key fields (because `cover` is a weird name for a team member's photo) AND multiple instances of the upload widget for different fields in the same object

czaja307:
yeah, i think that we can rename the fields without hesitation (@mini-bomba side note: there will be somen changes to science club in directus so our decision about going into profuction mode was a little bit premature)

Leadman5555:
Actually, that might not be needed. I'll think of a way to make it look rather clean and support whatever names.

Leadman5555:
- Added support for any property and any numbers of properties. 
- Added a way to remove the current image, and a flag to disallow removing current images (for version screenshots, where image is required). 
- Added margin
- I have no control about what appears first or the label text. I tried to make it more user friendly, but this... framework automatically sorts the fields and formats the label text for checkboxes. Similarly, it's impossible to show a property on edit, but hide on new - it appears to be the same action under the hood.

image

mini-bomba:
tbh I can tell from your screenshot alone that a checkbox to delete an image is gonna be jank, and it won't be intuitive...
so I have a suggestion to consider:
what if you tried to modify the `PhotoDropbox` component to handle everything about the image: 
- showing an immutable preview in the `show` view
- allowing for a new file to be uploaded in the `edit` view it none exists yet
- showing a preview of the current image with options to remove or replace it if one exists

this might require a bit more effort, but I think it's a better approach than trying to hack it in by only modifying the resource definition

Leadman5555:
The preview in the show / edit view is doable.  
As for the checkbox, I disagree with you on this one:
 - checked → deletes the current photo
 - unchecked → keeps it as unless they upload a new image to replace it  
 
 Use cases:
 - remove the current photo → check the box
 - upsert current photo → upload a new photo

It's impossible to break it, because if checked, it doesn't matter if you upload a new image or not - it's ignored anyway.
If we move all logic into the `PhotoDropbox` component, not only will we lose the all functionality provided by AdminJS, but also make it more janky - especially if it is supposed to react to different view modes... All things considered, it's the admin panel and not something that any user will interact with, so making it intuitive isn't really a priority here - using a tested framework API is. None of us with access to the admin panel would be your worst, completely clueless end-user after all. We would also decouple updating the photo attribute(s) with updating everything else, which doesn't seem like a good idea either, especially if rollbacks were ever needed due to validation errors. And in rare cases where photos are required, the component logic would have to be changed based on information from custom params from AdminJS state, which sounds... well, not the best, at least, in my opinion, that is. So, let us leave it as it is. The preview is a great idea, though - I'll implement it tomorrow.

mini-bomba:
> It's impossible to break it

That's not my point. The interface itself is jank and unintuitive from the user's perpective.
As a user, I expect there to be one widget per image property, and I expect it to be in a single state at any time out of a few possible ones: there was no image and none is being uploaded, there's an image (new, existing or replaced), there was an image and it's being deleted

> if checked, it doesn't matter if you upload a new image or not - it's ignored anyway.

that's even worse from the user's perspective: you're allowing the frontend to be in an invalid state and you're fixing it in some way on the backend without communicating to the user what is being done

> If we move all logic into the PhotoDropbox component, not only will we lose the all functionality provided by AdminJS, but also make it more janky - especially if it is supposed to react to different view modes

AdminJS doesn't appear to be providing any functionality for image props anyway (otherwise I'd expect you to be using it already?)
I feel like having a single adminjs property with a properly implemented widget would be a cleaner solution (and thus not a jank) as what you're doing currently.
Also let me be clear: I'm not asking you to implement *all* logic in the frontend component (that's impossible, you don't have access to backend utilities from the frontend, including the `FileService` which doesn't even have a production upload API at this moment)
I'm suggesting that you implement all possible state transition intentions (i.e. upload/replace/delete) with basic validation (i.e. require a file upload if a file is required, don't show a delete button if it's required) on the frontend and then communicate that state back to the backend to handle.
You'll still need to use before/after hooks on the backend.

> All things considered, it's the admin panel and not something that any user will interact with, so making it intuitive isn't really a priority here
> None of us with access to the admin panel would be your worst, completely clueless end-user after all.

I'm following the same policy as I do with error handling: I treat other developers (including future me) mostly the same as I'd treat a regular user.
I don't expect anyone to be familiar with every nook and cranny of our codebase, and neither do I expect them to be familiar with all the possible edge cases and gotchas of our code.
I want any errors and invalid states to be communicated clearly, in the interface (whether that be graphical UI or code), the inline code documentation, and the errors.
I don't want anyone (including future me) to waste time debugging an issue or trying to remember how an unintuitive component worked.

Leadman5555:
Let me know if this is what you meant. As for the CSS, I think it looks all right, although I'm not a UI expert.
## Upload:
image
image
image
### Show
image
### New
image

Leadman5555:
> > It's impossible to break it
> 
> That's not my point. The interface itself is jank and unintuitive from the user's perpective. As a user, I expect there to be one widget per image property, and I expect it to be in a single state at any time out of a few possible ones: there was no image and none is being uploaded, there's an image (new, existing or replaced), there was an image and it's being deleted
> 
> > if checked, it doesn't matter if you upload a new image or not - it's ignored anyway.
> 
> that's even worse from the user's perspective: you're allowing the frontend to be in an invalid state and you're fixing it in some way on the backend without communicating to the user what is being done
> 
> > If we move all logic into the PhotoDropbox component, not only will we lose the all functionality provided by AdminJS, but also make it more janky - especially if it is supposed to react to different view modes
> 
> AdminJS doesn't appear to be providing any functionality for image props anyway (otherwise I'd expect you to be using it already?) I feel like having a single adminjs property with a properly implemented widget would be a cleaner solution (and thus not as jank) than what you're doing currently. Also let me be clear: I'm not asking you to implement _all_ logic in the frontend component (that's impossible, you don't have access to backend utilities from the frontend, including the `FileService` which doesn't even have a production upload API at this moment) I'm suggesting that you implement all possible state transition intentions (i.e. upload/replace/delete) with basic validation (i.e. require a file upload if a file is required, don't show a delete button if it's required) on the frontend and then communicate that state back to the backend to handle. You'll still need to use before/after hooks on the backend.
> 
> > All things considered, it's the admin panel and not something that any user will interact with, so making it intuitive isn't really a priority here
> > None of us with access to the admin panel would be your worst, completely clueless end-user after all.
> 
> I'm following the same policy as I do with error handling: I treat other developers (including future me) mostly the same as I'd treat a regular user. I don't expect anyone to be familiar with every nook and cranny of our codebase, and neither do I expect them to be familiar with all the possible edge cases and gotchas of our code. I want any errors and invalid states to be communicated clearly, in the interface (whether that be graphical UI or code), the inline code documentation, and the errors. I don't want anyone (including future me) to waste time debugging an issue or trying to remember how an unintuitive component worked.

There has been a misunderstanding, I thought you wanted me to handle uploading images just in the photoDropbox component and make requests to some new endpoint from it. That's why I wrote about leaving AdminJS functionalities out

Leadman5555:
As for the failed checks, that's on the main branch. 
image

mini-bomba:
That's an improvement and it is closer to what I had in mind.
The upload field could be hidden behind a button if an image exists already, but this is already good.

The linter passes without issues on the main branch, so the errors are likely caused by changes on your branch (likely the `tsconfig.json` change, if I had to guess)

Leadman5555:
I've started checking it. You are right, removing DOM lib fixes it. Although it breaks the ES-lint in my part. I'll try to figure out why

mini-bomba:
You'll probably need to find a way to set TS compiler options only for the frontend components directory

Leadman5555:
I tried to fix it for the past two hours or so, and there are quite a few problems, the main of which is:  
We are building with `node ace build` which does not pick up references in tsconfig files. To build with two config files, we would need to either build the whole project with Adonis builder and then separately build the frontend components using different tsconfig file or use a different builder all together.   
Overall, it deserves its own issue. I pushed a commit with ts-expect-error that allows the build to go through for now.

mini-bomba (approved):
I've decided to push some "minor" changes directly to the PR branch.

Most notably, I've rewritten most of the PhotoDropbox component to have some better UI/UX (in my opinion)

https://github.com/user-attachments/assets/94819c77-b24a-4705-88fa-66cf54b4ca88

czaja has approved tests of this PR in production, :fire: :shipit:

PR #132, opened 2025-03-20
refactor: unify error handling

### LINKED ISSUE #126: refactor: unify error responses

mini-bomba:
stuff done in this PR:
- created a new error handling framework:
  - context stacks my beloved
  - defined an interface specifying how errors should be structured (`IBaseError`)
  - created a convenience class that allows values from that spec to be easily set (`BaseError`, though we could just be throwing plain JS objects around and it'd still work)
  - defined how error responses should be structured
- redefined the adonis error handler
  - it now uses the new error handling framework to parse error stacks
  - generates error responses as specified by the error framework
  - creates detailed error logs for non-silent errors
- implemented a rust-inspired method for easily wrapping errors
  - `.addErrorContext()` on `Promise`s
  - involves prototype modification
  - only works for async functions (but it's still useful)
  - requires a side-effect import to work
    - technically we should be placing these in each file that makes use of this function, but TS-LS doesn't do it automatically, so i just placed it in `start/kernel.ts` so that it should be loaded by the time the HTTP server starts
    - rust does it better, traits my beloved
 
closes #126 

som screenshots:

<details>
<summary>
`.addErrorContext()` in action
</summary>

image
</details>

<details>
<summary>
sample response for a vine validation error
</summary>

image
</details>

<details>
<summary>
sample response for 404 not found error (due to .findOrFail(), wrapped with .addErrorContext())
</summary>

image
</details>

<details>
<summary>
response for a simulated FileService failure
</summary>

image
</details>

<details>
<summary>
logs for a simulated FileService failure
</summary>

image
</details>


czaja307 (approved):
absolute fire

Just one concern/question - do we want to serve users with such detailed error description? I'm wondering if logging errors and returning just the general description wouldn't be a better approach? (we could also use some debug flag in env). However, if you think this is the way to go feel free to :shipit:

mini-bomba:
I don't think it's harmful to include the full context stack in most errors.
- At worst it'll be ignored by the client. (the top-level/most recent message is sent in a separate field anyway)
- At best, it can help detect a potential bug where we wrap an unexpected error as if it was expected and the client's fault.
  - In that case, the context stack would include the message of the unexpected error, which would be visible if someone would debug requests against the production server. 
  - If we omitted the context stack from responses, you'd need to set up a debug server instance to debug the issue, as 4xx errors are omitted from logs by default.
- For sensitive endpoints/errors, the errors can be marked as `sensitive` at any error layer and the context stack will be omitted from the response.

Additionally, the stack traces visible in screenshots are only sent if the server is in debug mode. The production server will not be including that field in responses.

PR #140, opened 2025-03-31
fix(admin panel): objects with a luxon date field cannot be saved

czaja307:
seems to be working now


czaja307:
it would've been to easy 🙃

mini-bomba (approved):
:fire: :shipit:

PR #143, opened 2025-03-31
ci: better migration tests

mini-bomba:
This PR adds various new scenarios & tests to the migration test workflow.

- in the existing migration test job: 
  - added a step to run seeders
- new jobs:
  - Run migrations from the base branch, then run migrations from the PR branch, then seed
    - aims to detect changes that break the models after an incremental migration
  - Run migrations from the base branch, then seed, then run migrations from the PR branch
    - aims to detect migrations that break if the database contains existing data
  - Source-only check/lint: verify ordering of new migration files
  - Run migrations from the base branch on one database, then migrations from the PR branch on another (excluding any new migrations), then diff the two databases
    - aims to detect intentional and accidental changes to existing migration behaviour

### DIFF
--- a/.github/workflows/db.yml
+++ b/.github/workflows/db.yml
@@ -7,7 +7,7 @@ on:
     branches: ["main"]
 
 jobs:
-  migration-check:
+  fresh-db:
     runs-on: ubuntu-latest
     env:
       DB_HOST: 127.0.0.1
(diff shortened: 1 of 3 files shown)


czaja307 (approved):
:fire: :shipit:

PR #144, opened 2025-04-02
refactor(models): make files db objects with relations

### LINKED ISSUE #135: feat: return path to files instead of hash

mini-bomba:
closes #135 

- all file key fields now end with _key
- all objects that have a file key field also have a file relation
  - this relation can be requested via the api to get a full file url


mini-bomba:
1 test failure due to the seeders from the main branch inserting invalid data

czaja307 (dismissed):
🔥 :shipit:

PR #146, opened 2025-04-06
feat(auth): add api token auth

czaja307:


start/routes.ts
czaja307, line 1: question: @mini-bomba how do you think, we should handle leaving GET requests for resource unsecured?

mini-bomba: imo there's no real reason to restrict data retrieval
- the existing cms system does not require auth to retrieve data
- we aren't storing any confidential data that we must hide from everyone where possible
  - and if we are, then only the endpoints that return these kinds of data should require auth
- the mobile team will need access to these endpoints from the app
  - they would need to ship an auth token in the app
    - while they can obfuscate shipped secrets, obfuscation is security through obscurity and security through obscurity is ineffective
      - a determined attacker would be able to extract the token anyway
  - this would just add unnecessary complexity on their side

czaja307: i might have misspoken - i wanted to ask what is the most efficient way to put auth APART FROM get requests

mini-bomba: since we've got the auth middleware... we probably should just use the auth middleware?
this would make it easy to forget to add auth to some endpoint tho :/
(i might look into autogenerating routes for the automatic crud tho - that would eliminate this issue)

mini-bomba: yeh but i don't think you can exclude a single route from having middleware inherited from the group applied to it

mini-bomba (commented):
imma add just a few changes from myself and we can :fire: :shipit: to prod

mini-bomba (approved):
:fire: :shipit:

PR #147, opened 2025-04-07
refactor(models): refactor typedModel into typedColumn

mini-bomba:
this is yet another big refactor on my way to automatic crud handling

this time, I took the existing `@typedModel` decorator and refactored it into the new `@typedColumn` decorators
- the old `@typedModel` decorator was one written from scratch, that had to be called on the model, with field names repeated as an input
  - this made it prone to errors during model refactors and changes
- the new `@typedColumn` decorators are basically an extension of adonis' `@column` decorators
  - they completely replace adonis' `@column` decorators and function like those decorators
  - they can take the same inputs as their adonis counterparts, with extra added parameters specific to typing
    - the mandatory* `type` parameter is mostly the same as the record values in `@typedModel`
      - `DateTime` cannot be specified manually on regular `@typedColumn` decorators
        - this is to encourage the use of `@typedColumn.date` and `@typedColumn.datetime` which are the correct decorators to use for datetime fields
      - there are a few custom types that map to our existing types, but produce different default validators
        - `integer` - maps to `number`, produces a default validator that only accepts integers
        - `uuid` - maps to `string`, produces a default validator that only accepts uuid4s
      - *the `type` field is only mandatory on regular `@typedColumn` - it is unavailable on `@typedColumn.date[time]`
    - the new `validator` param, which can be used to override the default validators generated for the given field
      - except if the type is an enum
    - `optional`, which allows the field to be marked as optional (aka nullable & undefinedable)
    - `autoGenerated`, which can be used to explicitly mark the field as auto generated (aka users should not be able to manually modify it)
      - a column is implicitly considered as `autoGenerated` when either `primaryKey` or `autoCreate` is set to true
  - our custom meta values for columns now store the new `optional`, `validator` and `autoGenerated` params
    - this will be used by the automatic crud handlers to generate appropriate validators for requests
- because the new decorator extends `@column`, it must be applied on each column instead of once per model
  - this approach makes it less prone to errors (like typos, or forgetting to update the `@typedModel` declarations when renaming fields)
  - you no longer have to write the same column name in two places or remember to update them

i also added typings to all models that I expect to be modifiable via the API, in preparation for that automatic crud

stuff to watch out for during review:
- please make sure I didn't mess up the declared types
  - it's quite likely that I might've used a base type (such as `number`) where a custom type should be used instead (such as `integer` on primary/foreign keys)

this pr was developed during two train trips


czaja307 (approved):
:fire: :shipit: 
#testujemynaprodukcji

PR #152, opened 2025-04-07
fix(typedColumn): remove inconsistencies

karolwolek:
I have tested the last #147 pr that refactored the typedModel decorator and i noticed some inconsistencies.
- Some of the validators were not properly injected into model `meta` property and some were
  - it was because they were assigned to the `options.validator` (which was a property to assign manual validators where decorator was used) instead of `options.meta.validator)
- At the beginning of the decorators functions there is redundant type intersection as `MainTypedColumnOptions` and `DateTypedColumnOptions` already have `Partial<ColumnDef>` merged in type declaration.
-  `version` and `regular_hours` had improper `optional` and `type` assigned in one of the columns


mini-bomba (approved):
:heart: :shipit:

PR #155, opened 2025-04-11
fix(admin panel): date picker timezone 

Leadman5555:
Visually works exactly like before.

### DIFF
--- a/app/admin/resources/resource_factory.ts
+++ b/app/admin/resources/resource_factory.ts
@@ -251,10 +251,8 @@ export class ResourceFactory {
     }
   }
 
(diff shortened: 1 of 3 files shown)


app/admin/resources/resource_factory.ts
mini-bomba, line 1: :fire:
just one question tho:
have you tried overwriting the autogenerated property instead of creating our own?

Leadman5555: That is... actually a good idea. New simplified version available

czaja307 (approved):
:fire: :shipit:

PR #163, opened 2025-04-17
feat(departments): added link names to departmentLinks

Leadman5555:
Works with scrapers, seeders, admin panel, and endpoints. Called the field `name` to keep the convention from Directus


czaja307 (approved):
:fire: :shipit:

PR #164, opened 2025-04-22
feat: automatic CRUD handling

### LINKED ISSUE #133: feat: add all CRUD endpoints

mini-bomba:
finally closes #133 
this PR took way too long
and as a result is way too big
(i apologize to anyone actually reviewing this)

## overview of changes
- added a `@typedManyToMany` decorator for typing pivot tables on n:m relations
- removed the `model` parameters from all scopes, in favour of extracting the model type from the query (`query.model`)
- defined all CRUD endpoints on base controller
  - `POST /` with all required properties: store a new object
  - `DELETE /:id`: deletes an object
  - `PATCH /:id` with any subset of properties: updates an object
  - `GET /:id:/:relation`: lists all related objects of the requested object (only relations from `crudRelations`, all existing scopes work)
  - `POST /:id:/:relation` for `hasMany` relations, with all required properties of the related object (except the foreign key for that relation): creates a new related object and associates it with the requested object
  - `POST /:main_id/:relation/:related_id` for `manyToMany` relations, with all required pivot properties: attaches the related object to the requested object
  - `DELETE /:main_id/:relation/:related_id` for `manyToMany` relations, with any subset of pivot properties marked as `detachFilter`: removes all matching attachments
- created automatic route definitions for BaseController-based controllers
  - disabled hot reloads for controllers because adonis does not support dynamically changing routes on hot reload :/
- autogen code got moved out of the base controller, into utils
- created a foreign key vine validator
  - cleanly validates whether a field is a valid foreign key
  - automatically copies the validator of the foreign key, and then queries the database to verify that an object exists
- added a new `foreignKeyOf` option to typing decorators
  - replaces manually entering a type and validator
  - takes a function that resolves to a model, to avoid issues with circular dependencies
    - this required some hackery to lazily compute model typing without requiring other functions to call functions or similar
    - as a result, all typing fields got moved to `meta.typing`
- created missing controllers: now all models have a corresponding controller
- as a smol bonus: renamed the `departments_links` table to `department_links` (cuz it annoyed me)

## tips for reviewing
- if the changed line count is too intimidating at first, try reviewing commit by commit
  - i tried to split them up where possible, but that one big (1k changes) commit was simply unavoidable due to how my coding sessions were spaced out in time
  - or maybe just take a look at the files in the tree, outside of the diffs (large diffs of refactors can be noisy and make it difficult to read the actual code)
- verify that all endpoints that should require auth actually require auth
  - i found this kind of issue just before creating the PR, during my final look through the diffs
- try to fetch something from each controller
  - just to verify that all self validation checks pass
  - (this is a great candidate for an automatic test if anyone's bored)
- verify that all types of endpoints actually work
  - i did a few tests while developing, but I might've missed something
  - since the code is highly generic, testing on one or two controllers should be enough, as long as you cover all types of handler functions


czaja307 (approved):
:fire::shipit:
let's test on prod (which isn't prod apparently)

PR #166, opened 2025-05-01
Feature(admin): relations (base, one-to-many)

Leadman5555:


Leadman5555:
Oh, we need to share the license key for the checks to pass? What do we do about it?

mini-bomba:
modify the workflows to pull the license key value from secrets

Leadman5555:
I don't think I have access to GitHub secrets on this repo (https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions), and as such I cannot add the secret myself

mini-bomba:
I already added the secret for you, just modify the workflows to use that value

app/models/change.ts
```diff
@@ -40,6 +40,10 @@ export default class Change extends BaseModel {
   @belongsTo(() => Version)
   declare version: BelongsTo<typeof Version>;
 
+  public static getVersionRelationKey() {
+    return "versionId";
+  }
```
czaja307, line 45: issue: I'm not a big fan of defining relation keys like this. 

This solution add unnecessary dependencies between models and adminjs resources. I don't see any benefit in passing hardcoded string via method instead just doing it in the resource directly.

czaja307 (changes requested):
:fire: the code looks solid, just a few minor issues

additional request: can you also add  the reverse relations using `reference`? I'm sending you how I did that in translator repo.
```tsx
originalLanguageCode: {
        reference: "languages",
},
```
https://github.com/Solvro/backend-translator/pull/12/files

czaja307 (approved):
:fire: :shipit: 

good job, sorry that it was you who had to fix all the minor issues around it

PR #127, opened 2025-03-15
feat(admin): add validation and error feedback

kamilfedio:
> [!IMPORTANT]
> This pull request adds validation for admin resources using `vine`, integrating it into resource actions to ensure data integrity before creating or editing records.
> 
>   - **Validation**:
>     - Adds `validateResource` function in `validators/utils.ts` to handle validation using `vine`.
>     - Integrates validation in `new` and `edit` actions for resources in `about_us.ts`, `academic_calendars.ts`, `admin_panel.ts`, `buildings.ts`, `departments.ts`, `guides.ts`, `student_organizations.ts`, and `versions.ts`.
>   - **Validators**:
>     - Adds new validator files: `about_us.ts`, `academic_calendars.ts`, `admin_panel.ts`, `buildings.ts`, `departments.ts`, `guide.ts`, `student_organizations.ts`, `versions.ts`.
>     - Each validator file defines specific validation rules for its respective resource model using `vine`.
>   - **Misc**:
>     - Imports `ActionRequest` in resource files to support validation integration.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=Solvro%2Fbackend-topwr&utm_source=github&utm_medium=referral)<sup> for 2771d9513cb058a962efcb22ac95b6e7a1bdd6b4. It will automatically update as commits are pushed.</sup>

PR #168, opened 2025-05-05
feat(admin_panel): validate requests

### LINKED ISSUE #92: feat(admin panel): add error handling
we should valiate the data in the panel if we can, if not we should find the way to show user more verbose errors
Image

mini-bomba:
closes #92 
supersedes aka closes #127 

this is an adaptation of kamilfedio's #127 PR to the new frameworks built up over 3 major refactors
some inspiration was taken from the original PR, but many things have been done differently, as the refactors have created simplified ways to achieve the same things - the original author is credited as a co-author where applicable
most notably, I've adapted some of the original validator rules, as they were in some cases more detailed than what we have currently

as I've simply reused the existing model typing framework in this PR, the changes also automatically affect automatically generated CRUD endpoints in exactly the same way.
and the opposite is also true - this PR now automatically benefits from all the foreign key validation improvements made for automatic CRUD

<details>
<summary>Some screenshots</summary>

image
image

</details>


czaja307 (approved):
:fire::shipit:

PR #177, opened 2025-05-19
feat: add pink boxes scraper

AlicjaBonar:
I created a new scraper that reads pink box locations from a local JSON file, validates input structure and creates PinkBox entries in the database.

### DIFF
--- a/database/scrapers/pink_boxes.ts
+++ b/database/scrapers/pink_boxes.ts
@@ -52,9 +52,7 @@ export default class PinkBoxScraper extends BaseScraperModule {
         }
         return data;
       });
-    for (const data of pinkBoxesData.data) {
-      const pinkBox = await PinkBox.create(data);
-    }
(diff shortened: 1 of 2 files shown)


database/scrapers/pink_boxes.ts
```diff
@@ -0,0 +1,60 @@
...
+          throw new Error(`
+                    Invalid JSON structure in ./assets/pink_boxes.json,
+                    expected type of Response<PinkBoxDraft>
+                    `);
+        }
+        return data;
+      });
+    for (const data of pinkBoxesData.data) {
+      const pinkBox = await PinkBox.create(data);
+    }
```
mini-bomba, line 57: consider passing the whole data array into `PinkBox.createMany` instead

AlicjaBonar:
All I changed was passing the whole data array into PinkBox.createMany instead of separately as you've suggested.

mini-bomba (approved):
guess removing the fields works too

PR #178, opened 2025-05-20
feat: cache reference number

karolwolek:
Added cache reference number singleton with scraper and adminjs integration. 

I also added the custom action for refreshing (bumping by one) with a guard - double confirmation.
After the pr is accepted i will add localization for the guard


database/migrations/1747747312692_create_cache_reference_numbers_table.ts
```diff
@@ -0,0 +1,23 @@
...
+export default class extends BaseSchema {
+  protected tableName = "cache_reference_numbers";
+
+  async up() {
+    this.schema.createTable(this.tableName, (table) => {
+      table.increments("id");
+      table.integer("reference_number").unsigned().notNullable();
+
+      table.timestamp("created_at");
+      table.timestamp("updated_at");
```
mini-bomba, line 12: make these not nullable

app/models/cache_reference_number.ts
```diff
@@ -0,0 +1,19 @@
...
+import { BaseModel } from "@adonisjs/lucid/orm";
+
+import { typedColumn } from "#decorators/typed_model";
+
+export default class CacheReferenceNumber extends BaseModel {
+  @typedColumn({ isPrimary: true, type: "number" })
+  declare id: number;
+
+  @typedColumn({ type: "number" })
+  declare referenceNumber: number;
```
mini-bomba, line 12: use the more specific "integer" type for these

app/controllers/cache_reference_number_controller.ts
mini-bomba, line 1: consider adding an endpoint to bump the ref number

mini-bomba (changes requested):
minor issues

app/controllers/cache_reference_number_controller.ts
czaja307: generally a good idea, but i don't think that we need it

mini-bomba: won't the frontend project need to bump the ref after approving changes?
or are we handling change approval with our adminjs panel?

czaja307: okay, good point, forgot about frontend :sob:

czaja307 (changes requested):
LGTM, just one minor improvement suggestion :fire:

app/controllers/cache_reference_number_controller.ts
```diff
@@ -0,0 +1,31 @@
...
+    await db.transaction(
+      async (trx) => {
+        const crn = await CacheReferenceNumber.query({
+          client: trx,
+        }).firstOrFail();
+        crn.referenceNumber++;
+        await crn.save();
+      },
+      { isolationLevel: "read committed" },
+    );
```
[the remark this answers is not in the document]
karolwolek: Turn's out there is an easy way to increment in lucid even without raw query

simon-the-shark:
what's the status?

czaja307 (approved):
LGTM, 🔥 :shipit:

PR #183, opened 2025-05-29
feat: university newsfeed

Leadman5555:


app/controllers/newsfeed_controller.ts
```diff
@@ -0,0 +1,26 @@
...
+  async latest({ request, response }: HttpContext) {
+    await NewsfeedService.startNewsfeedUpdate();
+    const completeOnly = request.qs().completeOnly === "true";
+    const update = completeOnly
+      ? NewsfeedService.getLatestNewsfeedArticles(true)
+      : NewsfeedService.getLatestNewsfeedArticles();
+    if (update === null) {
+      return response.serviceUnavailable(
+        "Couldn't fetch newsfeed articles. Please try again later.",
+      );
```
mini-bomba, line 22: don't manually create error responses, throw an error with an appropriate status value instead

app/services/newsfeed_service.ts
mini-bomba, line 1: consider parsing the RSS feed instead, which should result in simpler and less fragile code
https://pwr.edu.pl/rss/pl/24.xml

mini-bomba (commented):
initial review (quite shallow)

PR #184, opened 2025-06-06
 fix(scraper): field of study semester count

Leadman5555:


database/migrations/1749227968774_alter_field_of_studies_and_add_study_types_table.ts
```diff
@@ -0,0 +1,71 @@
...
+export default class extends BaseSchema {
+  protected tableName = "fields_of_studies";
+  protected studiesTypes = ["1DEGREE", "2DEGREE", "LONG_CYCLE"];
+
+  async up() {
+    await this.schema.raw(`
+      DO $$ BEGIN
+        IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'studies_type') THEN
+          CREATE TYPE "studies_type" AS ENUM (${this.studiesTypes.map((type) => `'${type}'`).join(", ")});
+        END IF;
```
mini-bomba, line 12: what.
is this trying to detect if a type already exists?
we don't need that

database/migrations/1749227968774_alter_field_of_studies_and_add_study_types_table.ts
```diff
@@ -0,0 +1,71 @@
...
+      END $$;
+      DO $$ BEGIN
+        IF NOT EXISTS (
+            SELECT 1
+            FROM information_schema.columns
+            WHERE table_name = '${this.tableName}' AND column_name = 'studies_type'
+        ) THEN ALTER TABLE "${this.tableName}" ADD COLUMN "studies_type" "studies_type" NULL;
+        END IF;
```
mini-bomba, line 20: no, seriously - why are you trying to detect if a column already exists
it won't, we know the migrations that were created before this one
and if it does, then i'd rather it throw an error in CI instead of being silently ignored

database/migrations/1749227968774_alter_field_of_studies_and_add_study_types_table.ts
```diff
@@ -0,0 +1,71 @@
...
+      END $$;
+    `);
+    await this.db.rawQuery(`
+      UPDATE ${this.tableName}
+      SET studies_type = CASE
+                   WHEN semester_count = 12 THEN 'LONG_CYCLE'::studies_type
+                   WHEN is_2nd_degree = true THEN '2DEGREE'::studies_type
+                   ELSE '1DEGREE'::studies_type
+        END
+    `);
```
mini-bomba, line 30: I bet you need to throw this into a `defer` call, otherwise adonis will be adonis and will run this during dry runs
(i should make a CI workflow for ensuring that dry runs not modifying the db)

database/migrations/1749227968774_alter_field_of_studies_and_add_study_types_table.ts
```diff
@@ -0,0 +1,71 @@
...
+        ) THEN ALTER TABLE "${this.tableName}" ADD COLUMN "is_2nd_degree" boolean NULL;
+        END IF;
+        IF NOT EXISTS (
+            SELECT 1
+            FROM information_schema.columns
+            WHERE table_name = '${this.tableName}' AND column_name = 'semester_count'
+        ) THEN ALTER TABLE "${this.tableName}" ADD COLUMN "semester_count" INTEGER NULL;
+        END IF;
+      END $$;
+    `);
```
mini-bomba, line 54: same as above - we don't need these checks, just use standard adonis schema builders

database/migrations/1749227968774_alter_field_of_studies_and_add_study_types_table.ts
```diff
@@ -0,0 +1,71 @@
...
+    await this.db.rawQuery(`
+      UPDATE ${this.tableName}
+      SET is_2nd_degree  = (studies_type = '2DEGREE'::studies_type),
+          semester_count = CASE
+                             WHEN studies_type = 'LONG_CYCLE'::studies_type THEN 12
+                             WHEN studies_type = '2DEGREE'::studies_type THEN 4
+                             ELSE 7
+            END
+    `);
```
mini-bomba, line 63: needs to be in a `defer` call because adonis is adonis

database/scrapers/departments.ts
```diff
@@ -30,8 +31,7 @@ interface FieldOfStudyDraft {
   name: string;
   url: string | null;
   isEnglish: boolean;
-  is2ndDegree: boolean;
-  isLongCycleStudies: boolean;
+  studiesType: string;
```
mini-bomba, line 34: the `studiesType` field does not exist in the current directus schema

database/migrations/1749227968774_alter_field_of_studies_and_add_study_types_table.ts
Leadman5555: Same as above

database/scrapers/departments.ts
czaja307: it's been discussed with mobile team, it actually clarifies the logic
https://discord.com/channels/687360174377533442/1380578188757045318

database/migrations/1749227968774_alter_field_of_studies_and_add_study_types_table.ts
mini-bomba: see below; try running raw queries in `this.defer`

database/scrapers/departments.ts
mini-bomba: then first create and populate this field in directus, and then we can talk about merging this pr

database/migrations/1749227968774_alter_field_of_studies_and_add_study_types_table.ts
czaja307, line 1: @Leadman5555 could you explain to me now or if you prefer at weekly, why do we need this raw query and can't use lucid enum? i can't figure out why is this workaround needed 🤔

database/scrapers/departments.ts
czaja307: @Leadman5555 the interface describes the format of the source data so in this case we need to leave them unchanged

but from what i recall i think that you should be able to deduce the `studiesType` from the `is2ndDegree` and `isLongCycleStudies`

czaja307 (changes requested):

mini-bomba (commented):
:fire: 
just one minor issue

PR #186, opened 2025-06-09
refactor(buildings): make polinka a distinct map item type

AlicjaBonar:
Adds Polinka migration, model, controller, and a scraper that loads data from a local JSON file and uploads related images from https://admin.topwr.solvro.pl/assets


database/migrations/1748518887015_create_polinkas_table.ts
```diff
@@ -0,0 +1,64 @@
...
+      );
+
+      table.timestamp("created_at");
+      table.timestamp("updated_at");
+    });
+  }
+
+  async down() {
+    this.schema.dropTable(this.tableName);
+  }
```
mini-bomba, line 63: you forgot to drop the type here, which caused a check failure

app/models/polinka.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export default class Polinka extends BaseModel {
+  @typedColumn({ isPrimary: true, type: "integer" })
+  declare id: number;
+
+  @typedColumn({ type: "string" })
+  declare identifier: string;
+
+  @typedColumn({ type: PolinkaIcon })
+  declare iconType: PolinkaIcon;
```
mini-bomba, line 25: i think polinka will only ever have one icon type

database/scrapers/polinkas.ts
```diff
@@ -0,0 +1,113 @@
...
+
+interface PolinkaDraft {
+  name: string;
+  latitude: number;
+  longitude: number;
+  addres: string; //intentional typo
+  photoKey: string | null;
+  externalDigitalGuideIdOrURL: string;
+  externalDigitalGuideMode: ExternalDigitalGuideMode | null;
+}
```
mini-bomba, line 26: this type doesnt actually match what's in the json file

mini-bomba (changes requested):
minor issues

AlicjaBonar:
I fixed all the maintioned issues. Large amount of comits is because i did not know that git saves renamed files as new ones, so i had to fix it, but ill remeber about it in the future. I only did not edit the buildings scraper to prevent it from adding polinka as a building because i think that the easiest solution to it would be to remove them from the buildings source file that is stored on https://admin.topwr.solvro.pl/items/Buildings/ but as i have never i dont want to mess something up there

I’ve fixed all the mentioned issues. The large number of commits is caused by me initially not realizing that Git treats renamed files as new ones. I’ve fixed that, and I’ll remember about it from now on.

The only thing I didn’t change was the building scraper. I think that the simplest solution would be to remove Polinka entries directly from the source file at https://admin.topwr.solvro.pl/items/Buildings/.

database/scrapers/polinka_stations.ts
```diff
@@ -0,0 +1,113 @@
...
+            `);
+        }
+        return data;
+      });
+
+    for (const polinkaStation of polinkaStationsData.data) {
+      polinkaStation.photoKey = await this.semaphore.runTask(() =>
+        this.uploadCoverAndGetKey(polinkaStation),
+      );
+    }
```
mini-bomba, line 70: this doesnt need to be in a semaphore if you're awaiting the promise in the loop.
you'd need to use the semaphore if you used a function like `Promise.all`.

assets/polinkas.json
mini-bomba, line 1: unused duplicated file

mini-bomba (changes requested):
apart from these minor issues:
- you need to edit the buildings scraper to exclude polinka from the buildings table
- a recent change to the scraper framework added a new `shouldRun()` method - you should implement it once you update.

AlicjaBonar:
I believe everything is all right now, but let me know if there's anything else I should correct.

database/scrapers/buildings.ts
mini-bomba, line 1: the filtering you've done here is quite late - the polinka objects might cause a file to be uploaded which won't be used
consider filtering the data before it is used to make database or file system changes.

AlicjaBonar:
Thanks for catching those issues, I've just finished making the necessary improvements, hope it's okay now, but let me know if i should still change something more

mini-bomba (changes requested):
i just remembered that since you're adding a new object that references the `campus` model, you need to also add the appropriate reverse relation to that model.

looking solely at the diffs is such a bad habit for reviews

mini-bomba (approved):
the commit message of the last commit is a bit long, but thankfully that doesn't matter here
:fire: :shipit:

PR #189, opened 2025-06-12
feat: event calendar

Leadman5555:


czaja307 (commented):
i'd go with scrapper with scheduler (https://github.com/verful/adonis-scheduler)
it'd simplify the controller logic
@mini-bomba what do you think bout it?

app/controllers/event_calendar_controller.ts
```diff
@@ -0,0 +1,52 @@
...
+  typeof CalendarEvent
+> {
+  protected readonly queryRelations = [];
+  protected readonly crudRelations = [];
+  protected readonly model = CalendarEvent;
+
+  protected async storeHook(
+    ctx: CreateHookContext<typeof CalendarEvent>,
+  ): Promise<PartialModel<typeof CalendarEvent> | void | undefined> {
+    const event = ctx.request as CalendarEvent;
```
mini-bomba, line 20: incorrect type: this is not a `CalendarEvent` object, but a partial object containing only the columns of the model.
The correct typing of this variable would be `Partial<InstanceType<typeof CalendarEvent>>`, or `PartialModel<typeof CalendarEvent>`
It is safe to assume that non-autogenerated and non-optional fields will be present, but there isn't a good way to represent this with TS.

app/controllers/event_calendar_controller.ts
```diff
@@ -0,0 +1,52 @@
...
+    if (event.endTime.diff(event.startTime).milliseconds <= 0) {
+      throw new BadRequestException("End time must be after start time");
+    }
+    return undefined;
+  }
+
+  protected async updateHook(
+    ctx: HookContext<typeof CalendarEvent>,
+  ): Promise<PartialModel<typeof CalendarEvent> | void | undefined> {
+    const changes = ctx.request as Partial<CalendarEvent>;
```
mini-bomba, line 30: Same as above, minus the TS typing rant.

app/models/calendar_event.ts
```diff
@@ -0,0 +1,56 @@
...
+  })
+  declare startTime: DateTime;
+
+  @typedColumn.dateTime({
+    autoCreate: false,
+    autoUpdate: false,
+    autoGenerated: false,
+    validator: vine.luxonDateTime(),
+  })
+  declare endTime: DateTime;
```
mini-bomba, line 36: `vine.luxonDateTime()` is the new default validator for `@typedColumn.datetime` columns - passing it here is redundant (as are all other properties passed to the decorator tbh - those should all be defaults)

database/migrations/1749631863474_create_calendar_events_table.ts
```diff
@@ -0,0 +1,23 @@
...
+    this.schema.createTable(this.tableName, (table) => {
+      table.increments("id");
+      table.text("google_cal_id").nullable();
+      table.text("name").notNullable();
+      table.timestamp("start_time").notNullable();
+      table.timestamp("end_time").notNullable();
+      table.text("location").nullable();
+      table.text("description").nullable();
+      table.timestamp("created_at");
+      table.timestamp("updated_at");
```
mini-bomba, line 16: minor: specify these as explicitly `.notNullable()`

database/scrapers/event_calendar.ts
```diff
@@ -0,0 +1,214 @@
...
+    }
+  }
+
+  private static async removeOutdatedEvents() {
+    const bound = DateTime.now().minus({ days: FETCH_LIMIT_DAYS });
+    await CalendarEvent.query()
+      .where("start_time", "<", bound.toISO())
+      .whereNotNull("google_cal_id")
+      .delete();
+  }
```
mini-bomba, line 202: you probably should check the end time instead of start time, just in case there's some event lasting like a week or smth

database/scrapers/event_calendar.ts
```diff
@@ -0,0 +1,214 @@
...
+  async run(task: TaskHandle): Promise<void> {
+    try {
+      task.update("Updating event calendar.");
+      await EventCalendarUpdater.removeOutdatedEvents();
+      await EventCalendarUpdater.upsertDatabaseEvents(task);
+      task.update(`Event calendar updated.`);
+    } catch (error) {
+      task.error(error);
+    }
+  }
```
mini-bomba, line 213: don't handle errors manually here - the framework will do a much better job for you. (including printing the whole cause & call stack)
also `task.error` would need to be returned, it does nothing by itself :) (just yet another adonis moment)

database/scrapers/event_calendar.ts
```diff
@@ -0,0 +1,214 @@
...
+    const calendarResponse = await fetch(
+      `https://calendar.google.com/calendar/ical/${CALENDAR_ID}/public/basic.ics`,
+    );
+    if (!calendarResponse.ok) {
+      throw new Error(
+        `Failed to fetch calendar events - got response status code ${calendarResponse.status}`,
+      );
+    }
+    return calendarResponse.text();
+  }
```
mini-bomba, line 154: you reimplemented `this.fetchAndCheckStatus()` here

database/scrapers/event_calendar.ts
Leadman5555: Mhm. `task.update` logs the message, `task.error` is a return value. Ehh...

database/scrapers/event_calendar.ts
mini-bomba: then make it non-static?

app/controllers/event_calendar_controller.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  protected async storeHook(
+    ctx: CreateHookContext<typeof CalendarEvent>,
+  ): Promise<PartialModel<typeof CalendarEvent> | void | undefined> {
+    const event = ctx.request as Partial<InstanceType<typeof CalendarEvent>>;
+    if (event.startTime === undefined || event.endTime === undefined) {
+      throw new BadRequestException(
+        "Malformed request. Start and end time missing",
+      );
+    }
```
mini-bomba, line 25: This should already be caught by the autogenerated validators that run before this (but I guess you're just doing this to silence typescript?)

database/scrapers/event_calendar.ts
mini-bomba, line 1: i've also taken a deeper look into the event parser...
and my first reaction after seeing the .ics format is: couldn't we just parse this whole thing into json???
while I appreciate the effort you've put into this parser, there are a few too many magic values in there for my taste.
and also apparently the .ics format has a line length max, after which it just splits the line into a new one????? you don't seem to be handling that edge case correctly.

image
image

app/controllers/event_calendar_controller.ts
```diff
@@ -0,0 +1,57 @@
...
+  typeof CalendarEvent
+> {
+  protected readonly queryRelations = [];
+  protected readonly crudRelations = [];
+  protected readonly model = CalendarEvent;
+
+  protected async storeHook(
+    ctx: CreateHookContext<typeof CalendarEvent>,
+  ): Promise<PartialModel<typeof CalendarEvent> | void | undefined> {
+    const event = ctx.request as Partial<InstanceType<typeof CalendarEvent>>;
```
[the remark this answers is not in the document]
Leadman5555: I mean, we could just cast it as a CalendarEvent as it were in the beginning, even if it's not 100% accurate - we are not accessing any of the lucidModel functionality anyway.

app/controllers/event_calendar_controller.ts
Leadman5555: Precisely. The fields will always be defined as the object is already validated beforehand. That's why casting it to the calendarEvent type would be better, even though not fully accurate. No need to do pointless undefined checks. Alternatively, I could use the `!` operator and a comment to suppress the linter.

database/scrapers/event_calendar.ts
Leadman5555: Haven't seen the line splits - why would it work like that 😢? I'll use the parser from https://github.com/Solvro/backend-topwr/pull/195 after it gets merged and wrap the result and work with that

database/scrapers/event_calendar.ts
Leadman5555: Everything is static there, remnant of the previous service. Converting it to instance methods would be fine now as it has been moved to the scraper module

app/controllers/event_calendar_controller.ts
mini-bomba: Okay, sure - you've convinced me, you can cast it to a `CalendarEvent` or `ModelAttributes<CalendarEvent>`, without the `Partial<>`.

Leadman5555:
Adapted to the new parser. Tested and works just as it used to with the old one.

database/scrapers/event_calendar.ts
```diff
@@ -24,114 +25,137 @@ interface GoogleCalendarEventDto {
...
     if (
       !this.calendarString.startsWith("BEGIN:VCALENDAR") ||
       !this.calendarString.trimEnd().endsWith("END:VCALENDAR")
     ) {
       throw new Error("Calendar invalid");
     }
-    this.cursor = calendarString.length - 16;
+    const calendarRoot = parseICS(calendarString, (message) =>
+      task.update(message),
+    ).VCALENDAR;
```
mini-bomba, line 42: `this.logger.warn` is probably the better warning logger here... but you'll need to pass it down from the scraper module

database/scrapers/event_calendar.ts
```diff
@@ -0,0 +1,230 @@
...
+
+  //null - none left, undefined - malformed event
+  private extractLastEvent(): GoogleCalendarEventDto | null | undefined {
+    if (this.events.length === 0) {
+      return null;
+    }
+    const eventBlock = this.events.pop();
+    if (eventBlock === undefined) {
+      return undefined;
+    }
```
mini-bomba, line 69: `this.events.length === 0` and `this.events.pop() === undefined` are basically the same check

database/scrapers/event_calendar.ts
```diff
@@ -0,0 +1,230 @@
...
+  googleCalId: string;
+  startTime: DateTime<true>;
+  endTime: DateTime<true>;
+  name: string;
+  location: string | null;
+  description: string | null;
+}
+
+class CalendarParser {
+  private events: ICSValue[];
```
mini-bomba, line 28: The correct type of this value is actually `ICSObject[]`, but I only exposed `ICSObject` through the `ICSRoot` alias, as I didn't think it would be used. That's a bit silly on my part.

database/scrapers/event_calendar.ts
```diff
@@ -0,0 +1,230 @@
...
+    const eventBlock = this.events.pop();
+    if (eventBlock === undefined) {
+      return undefined;
+    }
+    if (typeof eventBlock === "string" || Array.isArray(eventBlock)) {
+      this.task.update(
+        "Invalid calendar event - block is not an ICSObject. Skipping...",
+      );
+      return undefined;
+    }
```
mini-bomba, line 75: The arrays in ICSObject properties should only have ICSObjects in them :thinking:

database/scrapers/event_calendar.ts
```diff
@@ -0,0 +1,230 @@
...
+    return this.parseEvent(eventBlock);
+  }
+
+  private static assertValidDateTime(
+    dateValue: ICSValue | undefined,
+  ): DateTime<true> | null {
```
mini-bomba, line 81: minor naming nitpick, but `assert` would usually throw an error
`extractDateTime`, maybe? `asDateTime`?

database/scrapers/event_calendar.ts
```diff
@@ -0,0 +1,230 @@
...
+    const startTime = CalendarParser.assertValidDateTime(
+      eventBlock[FIELD_LABELS.startDate],
+    );
+    const endTime = CalendarParser.assertValidDateTime(
+      eventBlock[FIELD_LABELS.endDate],
+    );
+    if (startTime === null || endTime === null) {
+      this.task.update(
+        `Invalid calendar event ${this.events.length + 1}, missing date labels. Skipping...`,
+      );
```
mini-bomba, line 125: the event index will definitely get out of sync here once you skip one event.
(also I accidentally made the indexes in errors/warnings from `parseICS` start at 1 :skull:)

database/scrapers/event_calendar.ts
```diff
@@ -0,0 +1,230 @@
...
+  ): Promise<GoogleCalendarEventDto[]> {
+    const calendarString = await this.fetchGoogleEvents();
+    const parser: CalendarParser = new CalendarParser(calendarString, task);
+    const bound = DateTime.now().minus({ days: FETCH_LIMIT_DAYS });
+    const events: GoogleCalendarEventDto[] = [];
+    let event = parser.getLastEvent();
+    while (event !== null && event.startTime >= bound) {
+      events.push(event);
+      event = parser.getLastEvent();
+    }
```
mini-bomba, line 187: I get the optimization you tried to make, but from what I can see, the events in the .ics file aren't actually sorted by start/end time - it looks more like a sort by creation time.
and you know - parseICS just parses the whole thing anyway, you might as well replace this quite ugly loop with a .map.filter chain.

database/scrapers/event_calendar.ts
```diff
@@ -0,0 +1,230 @@
...
+      const existingEvent = await CalendarEvent.findBy(
+        "google_cal_id",
+        event.googleCalId,
+      );
+      if (existingEvent !== null) {
+        existingEvent.startTime = event.startTime;
+        existingEvent.endTime = event.endTime;
+        existingEvent.name = event.name;
+        existingEvent.location = event.location;
+        existingEvent.description = event.description;
```
mini-bomba, line 206: `existingEvent.merge(event)`?

database/scrapers/event_calendar.ts
Leadman5555: I exported the `ICSObject` as well now

database/scrapers/event_calendar.ts
mini-bomba: oh right, i forgot that's the event queue, not the list of interpreted events.

database/scrapers/event_calendar.ts
```diff
@@ -0,0 +1,227 @@
...
+        const diff = a.endTime.diff(b.endTime).milliseconds;
+        if (diff !== 0) {
+          return diff;
+        }
+        return a.startTime.diff(b.startTime).milliseconds;
+      }); //sorted from earliest to last, with one most recent/in the future being last
+    for (let i = validEvents.length - 1; i >= 0; i--) {
+      yield validEvents[i];
+    }
+  }
```
mini-bomba, line 66: well that's a js feature I haven't seen used yet

database/scrapers/event_calendar.ts
```diff
@@ -0,0 +1,227 @@
...
+  //null - none left, undefined - malformed event
+  private extractEvent(
+    eventBlock: ICSObject,
+  ): GoogleCalendarEventDto | undefined {
+    if (Array.isArray(eventBlock)) {
+      this.task.update(
+        "Invalid calendar event - block is not an ICSObject. Skipping...",
+      );
+      return undefined;
+    }
```
mini-bomba, line 77: surprising that the linter doesn't pick this up as a condition that can never be met, since it's known to be an `ICSObject`, not an array.

database/scrapers/event_calendar.ts
```diff
@@ -0,0 +1,227 @@
...
+    return this.parseEvent(eventBlock);
+  }
```
mini-bomba, line 79: If you remove the unreachable condition, this function seems quite useless now

database/scrapers/event_calendar.ts
```diff
@@ -0,0 +1,227 @@
...
+      this.task.update(
+        "Invalid calendar event, missing googleCalId. Skipping...",
+      );
+      return undefined;
+    }
+    const name = CalendarParser.asString(eventBlock[FIELD_LABELS.name]);
+    if (name === null) {
+      this.task.update("Invalid calendar event, missing name. Skipping...");
+      return undefined;
+    }
```
mini-bomba, line 143: would be nice to still have the event indexes in these warnings
also `this.logger.warn` is a more better function for logging issues than `task.update`.
or the `logger` function you pass into the constructor...

database/scrapers/event_calendar.ts
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  private static async removeOutdatedEvents() {
+    const bound = DateTime.now().minus({ days: FETCH_LIMIT_DAYS });
+    await CalendarEvent.query()
+      .where("end_time", "<", bound.toISO())
+      .whereNotNull("google_cal_id")
+      .delete();
+  }
```
mini-bomba, line 219: If you're logging event insertions/updates, you might as well log the amount of events deleted.
I guess you might as well log the total amounts for insertions/updates instead of names - that would likely result in less time spent logging.

mini-bomba (commented):
final minor suggestions

database/scrapers/event_calendar.ts
Leadman5555: I divided the logging into two groups - `task.update` for main information (amount of added/deleted and such), `logger._` for details

Leadman5555:
I think we ignore it. Since they repeat, they will be in calendar every time either way, so every time we update the events, new repetitions will be fetched as well. In my opinion, there is little to no point in determining whether they repeat and, if so, how often do they repeat. Especially that our custom events do not support repeating ones.

mini-bomba:
oh and looks like we've caught up with the migration timestamps - that's cool

mini-bomba (approved):
lgtm, :fire: job
:shipit:

PR #192, opened 2025-06-14
feat(scrapers): conditional exectuion & module selection by cmdline

mini-bomba:
stuff done:
- improved error logging in scraper modules by using the existing error handling code
  - also improved the existing error handling code by trimming the current working directory from stack traces
    - this shortens the stack traces and makes them more readable
    - paths to own code are now relative to CWD, paths to external library code start with the library name
- bumped `@poppinss/cliui` to remove the unnecessary `@ts-expect-error` directive from scraper command code
- created a new method in the base scraper module: `shouldRun()`
  - by default always returns true
  - if false is returned, the module is skipped
  - this method was implemented for all existing modules - it simply checks whether the main tables that are to be edited are empty
- the modules to run can now be specified directly on the command line - just pass the names or file names as extra arguments
  - this makes it possible for some modules to be run non-interactively, as a scheduled task

### DIFF
--- a/commands/db_scrape.ts
+++ b/commands/db_scrape.ts
@@ -6,6 +6,7 @@ import path from "node:path";
 
 import { BaseCommand, args, flags } from "@adonisjs/core/ace";
 import type { CommandOptions } from "@adonisjs/core/types/ace";
+import { LucidModel } from "@adonisjs/lucid/types/model";
 
 import { LinkType, detectLinkType } from "#enums/link_type";
 import {
@@ -13,6 +14,7 @@ import {
   prepareReportForLogging,
   toIBaseError,
 } from "#exceptions/base_error";
+import { modelCount } from "#utils/db";
 
 /*
  * The scraper framework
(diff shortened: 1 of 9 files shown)


commands/db_scrape.ts
```diff
@@ -31,6 +38,8 @@ import { LinkType, detectLinkType } from "#enums/link_type";
  *   - the method may use `this.logger` for logging
  *   - the method should use `this.semaphore` for limiting the amount of parallel requests (if such are made)
  *   - the method may use utility methods inherited from the BaseScraperModule class
+ * - if your scraper script is expected to only run once, override the `shouldRun` method
```
czaja307, line 41: style: fix this indentation

commands/db_scrape.ts
```diff
@@ -163,6 +204,18 @@ export default class DbScrape extends BaseCommand {
   })
   declare maxJobs: number;
 
+  @flags.boolean({
+    alias: ["f"],
```
czaja307, line 208: isn't the f alias to dangerous? i'd go with full name only

czaja307 (commented):
:fire:
just a few minor issues below

commands/db_scrape.ts
mini-bomba: lol

PR #194, opened 2025-06-17
feat(validators): luxonDateTime validator model

mini-bomba:
vine pretends to be part of the adonis stack while not even supporting the datetime format adonis uses...
so i made my own clone of the date validator model that uses luxon datetimes instead

this is aimed at resolving some potential issues with date types, which may allow the amount of CRUD boilerplate in #189 to be reduced


Leadman5555:
LGTM 👍 
Side note: I've tested it a bit with the event calendar. To get the desired effect of the startTime and the endTime either being both null or both specified with startTime < endTime, you'd need this validator object to do the trick:
``` typescript
const testValidator = vine.object({
  startTime: vine.luxonDateTime().beforeField("endTime").nullable(),
  endTime: vine.luxonDateTime().afterField("startTime").nullable(),
});
```

Leadman5555 (dismissed):

PR #195, opened 2025-06-21
feat(utils): an ICS parser

mini-bomba:
this is a PR to the #189 PR.

it's basically a full ICS parser, although made to parse the google ICS docs, instead of referencing some actual documentation/spec/RFC.
feel free to use it to improve your existing scraper - this parser handles a few edge cases that I've noticed - split lines and inline objects.

some reference screenshots:
image
do note the `DTSTART: { VALUE: 'DATE:20180424' }` properties for some events - these appear to be full-day events
image


Leadman5555 (approved):
Overall, looks clean 👍 - use of a stack-like array is nice, haven't thought of it myself in my parser. 
Suggestion: I think we should flatten the date fields (with full days, that is, nested value) into a timestamp. Yes, it may make the parser less versatile (more specialized), but all Google calendars come in the same format, and the likelihood of us having to parse any other .isc file is next to none. Especially since the parser was designed with the Google ISC docs. Although, it's not really necessary, so let me know what you think about it.

mini-bomba:
The idea of this parser is to, well, parse the ICS into a more manageable format - JS objects.
It doesn't actually interpret the ICS data (such as converting various date(time) strings into luxon datetimes) - this is left up to you.

Leadman5555:
I didn't mean converting to ready-to-use types, but flattening the dates (`DTSTART: { VALUE: 'DATE:20180424' }` ->`DTSTART:  20180424T00:00:00`). But, as I've said, it's not needed as the result of the parser can be wrapped in another... parser.

PR #196, opened 2025-06-22
refactor(auth): use jwt instead

Batory12:
Auth now using jwt tokens. Send user info as payload.


mini-bomba (changes requested):
you've got check failures

also is this really the way to implement jwt? there's no adonis library for this?
this is copied from their docs, right?

Batory12:
well there is no official one afaik, I found [this one](https://packages.adonisjs.com/packages/adonisjs-jwt) if we wanna use some 3rd party

mini-bomba:
feels a bit silly on adonis' part
if the doc code actually works and isn't missing some major feature, then I guess we can go forward with that

app/auth/guards/jwt.ts
```diff
@@ -215,7 +215,7 @@ export class JwtGuard<UserProvider extends JwtUserProviderContract<unknown>>
   async authenticateAsClient(
     user: UserProvider[typeof symbols.PROVIDER_REAL_USER],
   ): Promise<AuthClientResponse> {
-    // @ts-ignore
+    // @ts-expect-error useleess
     const token = await this.generate(user);
```
mini-bomba, line 219: ????

czaja307:
> you've got check failures
> 
> also is this really the way to implement jwt? there's no adonis library for this? this is copied from their docs, right?

yeah that's weird (adonismoment) but the creators actually created docs section with tutorial how to implement custom JWT auth instead of adding it as option for the auth lib

czaja307:
@mini-bomba shouldn't we also implement the refresh token? seems like something we'd want to do here but i'd like to hear ur opinion about this

czaja307:
> well there is no official one afaik, I found [this one](https://packages.adonisjs.com/packages/adonisjs-jwt) if we wanna use some 3rd party

the lib seems okay at first glance (it's quite niche - but which adonis lib isn't)
i guess we could use it (unless @mini-bomba will veto me on this one) but the custom implementation is fine as well

app/auth/guards/jwt.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    /**
+     * Ensure the auth header exists
+     */
+    const authHeader = this.#ctx.request.header("authorization");
+    if (authHeader === undefined) {
+      // eslint-disable-next-line @typescript-eslint/only-throw-error
+      throw new errors.E_UNAUTHORIZED_ACCESS("Unauthorized access", {
+        guardDriverName: this.driverName,
+      });
```
czaja307, line 143: solvroconfig is deliberately quite strict to enforce good code quality and practices

disabling linter is only allowed in specific cases (linter is having a stroke/the alternative approach would need A LOT excessive code/the config rules are conflicting etc.) - if you have such a case here, write descriptive comments about it in code
otherwise just fix the errors instead of skipping them

czaja307 (changes requested):

Batory12:
ok, I fixed typing wherever possible

app/auth/guards/jwt.ts
```diff
@@ -0,0 +1,219 @@
...
+   * A flag to know if the authentication was an attempt
+   * during the current HTTP request
+   */
+  authenticationAttempted = false;
+
+  /**
+   * A boolean to know if the current request has
+   * been authenticated
+   */
+  isAuthenticated = false;
```
Leadman5555, line 89: Remember to set the flag after authentication

app/auth/guards/jwt.ts
```diff
@@ -0,0 +1,219 @@
...
+   * Generate a JWT token for a given user.
+   */
+  async generate(user: UserProvider[typeof symbols.PROVIDER_REAL_USER]) {
+    const providerUser = await this.#userProvider.createUserForGuard(user);
+    const { name, email, role } = user as {
+      name?: string;
+      email?: string;
+      role?: string;
+    };
+    const token = jwt.sign(
```
Leadman5555, line 106: Add other JWT fields:
- audience
- subject (that would be user email)
- issuer (that our server)
- expiry time
- issuedAt

app/auth/guards/jwt.ts
```diff
@@ -0,0 +1,219 @@
...
+      { userId: providerUser.getId(), name, email, role },
```
Leadman5555, line 107: We really don't need that in payload. Leave only role - there is no need to expose the userId at all and email will be JWT subject. Note that user email is unique - see the migration files for details.

app/auth/guards/jwt.ts
```diff
@@ -0,0 +1,219 @@
...
+     */
+    if (this.authenticationAttempted) {
+      return this.getUserOrFail();
+    }
+    this.authenticationAttempted = true;
+
+    /**
+     * Ensure the auth header exists
+     */
+    const authHeader = this.#ctx.request.header("authorization");
```
Leadman5555, line 137: `authorization` -> `Authorization` . It usually starts with a upper-case letter

Leadman5555 (changes requested):
Some much needed changes to implement - mainly filling out the JWT fields.
As for the `refresh-token`, we can generate it like the normal access tokens, but set the expiry time to a larger value. For example,  access token to 1 hour and refresh token to a longer period - look into the `rememberMe`  param in the login method (older version from main) - we can set the expiry of the refresh token based on the params value.
Additionally, we could use a different, more advanced algorithm, to sign the `refresh-token` - I'd suggest a fast one like HS256 for access tokens and an ES256 for the refresh ones (you can set the algorithm in the sign and verify method). Remember to use different keys for each of them.

Leadman5555:
And for the refresh token, add a `/refresh` endpoint to the `auth_controller.ts`, which verifies the refresh token (which is a cookie) and, if verification succeeds, sends back a new access token. We will send the detailed flow to the frontend team after we get the error responses in order and determine what is sent where and when 👍

app/auth/guards/jwt.ts
mini-bomba: http headers should be case insensitive

Leadman5555:
We can leave the Id in payload if we want to - not really needed, but having it there does not cause any errors; although it does make the token itself bigger. It cannot be used by anyone else than our server since only we can decode the token and on our server we can get the Id from the unique email. Unless there is a specific reason to leave it there (maybe @czaja307 knows more about it).
I'm not sure if I follow the rest of the message: 
- Without cookies meaning without refresh tokens?
- I have absolutely no idea what "dl" refers to or stands for

mini-bomba:
as for the refresh cookie: we've decided a long time ago that we do not want to use cookies in the API, as allowing cross-origin requests with cookies would open the API up to XSS attacks.
the refresh token idea can stay, but will instead be another token sent in response bodies, left for the client to manually save and later include in requests.

Leadman5555:
Assuming we ever allowed email change, then yes, that's a valid point. In that case, we can remove the email from the token altogether and leave just the Id instead. I still think sending both is redundant. 
As for the refresh tokens - if we discard the cookie idea, we need to save the refresh tokens in the database to know which ones are valid and to invalidate/expire them accordingly. It's not a worse approach by any means, just more code to write, I'd say. But since cookies are out, that's our way to go.

Leadman5555:
Seems so - it might have misread your comment a bit since it's late. Good that we agree on the fact that the cookies or not, we need to validate the expiry date (which the current implementation seems to be missing for the access tokens - although it has been marked as a resolved conversation a few messages above, so hopefully it has been included).

app/controllers/auth_controller.ts
```diff
@@ -1,34 +1,24 @@
...
-      expiresIn: rememberMe === true ? "30 days" : "1 day",
-    });
-
-    assert(token.value !== undefined, "Token value is missing");
 
-    return {
-      user,
-      token: token.value.release(),
-    };
+    return await auth.use("jwt").generate(user);
```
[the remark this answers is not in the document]
Leadman5555: As decided in the conversation, cookies will not be used

app/controllers/auth_controller.ts
```diff
@@ -1,34 +1,24 @@
...
 
   async me({ auth }: HttpContext) {
     return auth.getUserOrFail();
   }
 
-  async logout({ auth }: HttpContext) {
-    await auth.use().invalidateToken();
+  async logout() {
+    // Cant invalidate jwt, maybe do sth  later
+    // await auth.use().invalidateToken()
```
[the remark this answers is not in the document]
Leadman5555: As decided in the conversation, cookies will not be used

Leadman5555:
I removed most, well, nearly all of previous implementation. Sorry @Batory12. I just found it too convoluted and inefficient for no reason. 
I've tested the current version locally and it works as expected, both JWT validation and generation itself, and other endpoints, such as the `/refresh` or `/logout`.

Leadman5555 (dismissed):
LGTM 👍

czaja307 (dismissed):
LGTM - really neat solution :fire: 

Just one concern in the convo linked in the issue frontend requested the data about user to be passed inside the token - i dont see this (i may be blind because i've spent too much time doing wakacyjne wyzwanie reviews). If that's the case we can still merge this PR and add payload functionality after #225

Leadman5555:
The things the frontend requests are not possible due to the nature of the JWT tokens, because the token is encrypted and the keys are on our backend. They could have a partial key and read it, but that's not a correct approach. 
Additionally, storing unnecessary data in the JWT will slow the authentication as with each request the content needs to be decrypted - more content, more time spent on decrypting - and that's a very frequent operation.

I've just read the thread on DC and as @mini-bomba mentioned, there is a `/me` endpoint - they can get the information from there.

mini-bomba (approved):
:fire: :shipit:

PR #197, opened 2025-06-27
ci: quickfix command for PRs

Leadman5555:
Tested on a private repo and seems to work correctly (testing workflows with dummy PRs is a chore). 
How does it work:
 - type '/rerun-ci-please' in a PR comment
 - the `rerun-ci` workflow will be called
 - it adds a pending status check on the PR
 - calls the `ci` workflow
 - waits for up to five minutes for the `ci` workflow to complete
 - updates the pending status check to the result of the executed `ci` workflow

 
image

Script logs:

image

Actions view (first from the bottom is the cancelled PR commit check, second is the `rerun-ci` called by the comment, third is the new `ci` run):

image


mini-bomba:
You shouldn't need to rerun ci manually
Rerunning CI won't change anything anyway, as it's gonna be running on that old commit.

Leadman5555:
Not really, you can rerun it on the new one if you have the commit sha. But I guess it would be rerun on the new one automatically anyways. Then, since the rerun workflow is basically complete, I'll leave as an extra and add a second one that runs prettier (and maybe other stuff) and then commits

Leadman5555:
That's exactly what I've written - it would be rerun anyways.

> But I guess it would be rerun on the new one automatically anyways.

Leadman5555:
https://stackoverflow.com/questions/63188674/github-actions-detect-author-association
I've just read on the associations as well. I think we set it to COLLABORATOR?

As for the abuse, well, you can get rate limited on checkout action (I've gotten a few times today) 😆

mini-bomba:
never would've thought that you can get ratelimited by git cloning
anyway, again, the author association field is a single enum value, so you'll likely need to allow multiple values.
anyone who is a collaborator in this repo will likely also be a member, and member takes priority.
but i'm fine with allowing both member and collaborator to run this command. maybe even the PR author if you want.

mini-bomba:
you can just check whether the comment author == the pr/issue author - you get both in the webhook/ci payload
but this is optional

Leadman5555:
Added the quick-fix workflow - runs linter and prettier. Tested and works fine.

Before:
image

image


After:

image

image

mini-bomba:
yeah, it's a rare case
and you could also rebase & push to avoid empty commits - should have the same effect

mini-bomba (commented):
other than this one minor issue, :fire: :shipit:

.github/workflows/quick_fix.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Push changes if any made
+        run: |
+          git config --local user.email "github-actions[bot]@users.noreply.github.com"
+          git config --local user.name "github-actions[bot]"
+          git add .
+          HEAD_REF=${{ steps.pr.outputs.headRef }}
+          git diff --quiet && git diff --staged --quiet || (
+            git commit -m "fix: /quickfix command"
+            git push origin HEAD:$HEAD_REF
+          )
```
mini-bomba, line 64: didn't fix it - you need to pass in the environment variable using `env:`

mini-bomba (approved):
:fire: :shipit:

PR #200, opened 2025-06-30
refactor(buildings): delete building_icon

AlicjaBonar:
This pull request removes the icon_type column from the buildings table along with the associated enum type building_icon, as these are no longer needed. The changes are implemented through two separate migrations to avoid conflicts between dropping the column and removing the associated enum type. Otherwise, removing the enum type in the same migration caused errors, as the column referencing it had not been fully removed yet.


database/migrations/1751244539613_delete_building_icon.ts
mini-bomba, line 1: consider merging this migration with the previous one - this way you should be able to avoid the ugly manual type creation query

mini-bomba (changes requested):
- you've got check failures
- and you didn't make the same changes to models as you did to migrations
  - this must be done here, not in another PR, else we'll end up with a broken version of the backend

database/migrations/1751244539613_delete_building_icon.ts
```diff
@@ -0,0 +1,21 @@
...
+    await this.schema.raw(`
+      DO $$
+      BEGIN
+        IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'building_icon') THEN
+          CREATE TYPE building_icon AS ENUM ('ICON');
+        END IF;
+      END
+      $$;
+    `);
+  }
```
mini-bomba, line 20: also awaiting on these two is a mistake
just yet another adonis moment for you

database/migrations/1751244492545_alter_buildings_and_remove_building_icon.ts
mini-bomba, line 1: another PR with a migration got merged while this PR was in review - rename this file so that the timestamp is greater than `1751969302203`.
the current timestamp is `1752090041000`, which should work.

mini-bomba (changes requested):

mini-bomba (approved):
:fire: :shipit: once ci passes

PR #201, opened 2025-07-03
refactor(scrapers): make about us a scraper & deduplicate code

Leadman5555:


app/enums/link_type.ts
```diff
@@ -18,6 +18,12 @@ export enum LinkType {
   Twitch = "twitch",
 }
 
+export function stringToLinkType(value: string): LinkType | undefined {
+  return Object.values(LinkType).find(
+    (enumValue) => enumValue.toString() === value,
+  );
+}
```
mini-bomba, line 25: is this `toString()` actually needed?

commands/db_scrape.ts
```diff
@@ -46,6 +47,26 @@ const LOADABLE_EXTENSIONS = [".js", ".cjs", ".mjs", ".ts"];
...
+export function isValidDataResponse<T>(
+  response: unknown,
+): response is SourceResponse<T> {
+  return (
+    typeof response === "object" &&
+    response !== null &&
+    "data" in response &&
+    Array.isArray(response.data)
+  );
+}
```
mini-bomba, line 68: honestly this validator function is completely inadequate for what it's supposed to do

database/scrapers/about_us.ts
```diff
@@ -0,0 +1,143 @@
...
+        return data;
+      });
+    await AboutUsGeneralLink.createMany(
+      linkData.data.map((linkDraft) => {
+        return {
+          ...linkDraft,
+          linkType: stringToLinkType(linkDraft.linkType),
+        };
+      }),
+    );
```
mini-bomba, line 81: this `stringToLinkType` conversion function is purely to please TS, isn't it?
i'm sure the database wouldn't care.
(this might even make the potential database errors more cryptic, as you'll likely get a "required column missing error" instead of "this value is not a valid enum value", since your conversion functions converts invalid enum values to `undefined`)

database/scrapers/about_us.ts
```diff
@@ -0,0 +1,143 @@
...
+    const extension = (await this.fetchJSON(
+      filesMetaPath(file),
+      `file meta of ${file}`,
+    )) as { data: { filename_disk: string } };
+    try {
+      return extension.data.filename_disk.split(".").pop()?.toLowerCase();
+    } catch (e) {
+      throw new Error(`Failed to fetch extension for ${file}`, { cause: e });
+    }
+  }
```
mini-bomba, line 121: while you're already deduplicating scraper code - these functions are good candidates for moving to the base scraper class
if you do decide to do that - you can drop the "for model x" parts of error messages and add them at the callsite with `.addErrorContext()`

database/scrapers/about_us.ts
```diff
@@ -0,0 +1,143 @@
...
+function isValidDataResponse<T>(
+  response: unknown,
+): response is SourceResponse<T> {
+  return (
+    typeof response === "object" &&
+    response !== null &&
+    "data" in response &&
+    Array.isArray(response.data)
+  );
+}
```
mini-bomba, line 138: you just moved these to `db_scrape.ts` :thinking:

mini-bomba (commented):
initial review: a bunch of minor issues

app/enums/link_type.ts
Leadman5555: Yes, for typescript

commands/db_scrape.ts
```diff
@@ -46,6 +47,26 @@ const LOADABLE_EXTENSIONS = [".js", ".cjs", ".mjs", ".ts"];
 
 export type TaskHandle = Parameters<TaskCallback>[0];
 
+export function resolveDate(dateString: string): DateTime<true> {
+  const date = DateTime.fromISO(dateString);
+  return date.isValid ? date : DateTime.now();
+}
```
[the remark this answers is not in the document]
Leadman5555: I just copied those from buildings scraper. I'll refactor it

Leadman5555:
I have basically centralized the photo uploading logic. We had code that does the exact same thing written in four? different ways. Replaced them all with the new `baseScraper` method - `directusUploadFieldAndGetKey`, with different result handling depending on the use case

database/scrapers/about_us.ts
```diff
@@ -0,0 +1,83 @@
...
+    const formattedAboutUsData = {
+      description: draft.description,
+      coverPhotoKey: await this.semaphore.runTask(() =>
+        this.directusUploadFieldAndGetKey(draft.cover).then((value) => {
+          if (value === null) {
+            throw new Error("No cover photo provided");
+          }
+          return value;
+        }),
+      ),
```
mini-bomba, line 59: really minor: you don't have to use a semaphore if you aren't using `Promise.all`

database/scrapers/contributors.ts
```diff
@@ -142,26 +140,15 @@ export default class ContributorsScraper extends BaseScraperModule {
...
-            ).addErrorContext(() => `Failed to upload file ${id} to storage`);
-            return [id, file.id] as [string, string];
+            return [
+              id,
+              await this.directusUploadFieldAndGetKey(id).then((fileId) => {
+                if (fileId === null) {
+                  throw new Error(`Failed to upload file ${id}`);
+                }
+                return fileId;
+              }),
```
mini-bomba, line 150: what did you do to my beautiful error handling
this is why you don't return fake values as failures - just throw an error, you could've just used `.addErrorContext` and you would've preserved full cause information

database/scrapers/departments.ts
```diff
@@ -68,46 +67,34 @@ export default class DepartmentsScraper extends BaseScraperModule {
...
+                `https://admin.topwr.solvro.pl/items/${schema}?limit=-1`,
+                schema,
+              ),
+            )
+            .then((response) => {
+              assertResponseStructure(response);
+              return response;
+            }),
         ),
       )) as [
```
mini-bomba, line 82: you need to add error context here, else it will be impossible to tell which schema failed the check

mini-bomba (changes requested):

Leadman5555:
The reason why the upload function returns null instead of throwing is because some scrapers can accept null value (photo field is nullable), others cannot (required field). I guess I can flip the order, throw instead of null and use try/catch when null is accepted.

Leadman5555:
A model can have optional (nullable) photo key. It can be null because there was no photo or it can be null because, for any reason, the photo couldn't be obtained. If the model is valid, photo or not, then even if it tries and fails to get the photo, it can still insert the model without it.

mini-bomba:
If we're refactoring, we might as well improve the code.

database/scrapers/buildings.ts
```diff
@@ -73,16 +63,12 @@ export default class BuildingsScraper extends BaseScraperModule {
...
-    if (!isValidBuildingsData(buildingsData)) {
-      throw new Error(`
-        Invalid data type fetched from ${buildingsPath},
-        expected type of Response<BuildingDraft>
-        `);
-    }
+      "list of buildings from Directus",
+    ).then((response) =>
+      assertResponseStructure(response, "Buildings from Directus"),
+    )) as SourceResponse<BuildingDraft>;
```
mini-bomba, line 71: you can replace this promise chain with your new `.fetchDirectusJSON` method

database/scrapers/buildings.ts
```diff
@@ -96,13 +82,14 @@ export default class BuildingsScraper extends BaseScraperModule {
           latitude: data.latitude,
           longitude: data.longitude,
           haveFood: data.food ?? false,
-          coverKey: await this.semaphore.runTask(() =>
-            this.uploadCoverAndGetKey(data),
-          ),
+          coverKey:
+            data.cover === null
+              ? null
+              : await this.directusUploadFieldAndGetKey(data.cover),
```
mini-bomba, line 88: use `.addErrorContext` here to include something about the building here (either id or name), so we know which one failed.

database/scrapers/contributors.ts
```diff
@@ -142,26 +140,12 @@ export default class ContributorsScraper extends BaseScraperModule {
...
-            const file = await FilesService.uploadStream(
-              Readable.fromWeb(fileResponse.body),
-              extension,
-            ).addErrorContext(() => `Failed to upload file ${id} to storage`);
-            return [id, file.id] as [string, string];
+            return [
+              id,
+              await this.directusUploadFieldAndGetKey(id).addErrorContext(
+                "Images from Contributors imageIds",
+              ),
```
mini-bomba, line 147: Any string passed to `.addErrorContext` becomes the new top-level error message, so make sure it makes sense as an error message.

database/scrapers/departments.ts
```diff
@@ -69,45 +67,28 @@ export default class DepartmentsScraper extends BaseScraperModule {
       (await Promise.all(
         this.directusSchemas.map((schema) =>
           this.semaphore.runTask(() =>
-            this.fetchJSON(
+            this.fetchDirectusJSON(
               `https://admin.topwr.solvro.pl/items/${schema}?limit=-1`,
               schema,
-            ),
+            ).addErrorContext(`Failed for schema ${schema}`),
```
mini-bomba, line 73: same as above: the string should make sense as an error message - even if the context stack would reveal this, do mention *what* it actually failed to do (i.e. failed to download the schema)
also if the error context is generated at runtime (for example if it includes a template string) then it is preferred to wrap it in an arrow function, which will only be evaluated on error.

though as the fetch methods already include info about which schema failed (probably cuz they were made before i "borrowed" error handling schemes from rust) you may as well remove the error context, as it's redundant.

database/scrapers/departments.ts
```diff
@@ -69,45 +67,28 @@ export default class DepartmentsScraper extends BaseScraperModule {
...
-          `file ${departmentEntry.logo}`,
-        );
-
-        if (fileResponse.body === null) {
+        if (departmentEntry.logo === null) {
           throw new Error(
-            `Response body is null for department ${departmentEntry.id} with asset id ${departmentEntry.logo}`,
+            `Failed no logo for departmentEntry: ${departmentEntry.id}`,
           );
         }
```
mini-bomba, line 88: this error message is a bit weird

database/scrapers/faq.ts
```diff
@@ -167,22 +99,12 @@ export default class FaqSectionScrapper extends BaseScraperModule {
...
-        );
-      }
-
       await GuideArticle.create({
         id: article.id,
         title: article.name,
         shortDesc: article.short_description,
         description: article.description ?? "",
-        imageKey,
+        imageKey: await this.directusUploadFieldAndGetKey(article.cover),
```
mini-bomba, line 107: since this is being called in a loop, consider using `.addErrorContext` to include a new error message which mentions which article this image belonged to

database/scrapers/faq.ts
```diff
@@ -53,88 +42,31 @@ export default class FaqSectionScrapper extends BaseScraperModule {
...
+          "FAQ types",
+        ),
+        this.fetchDirectusJSON(
+          "https://admin.topwr.solvro.pl/items/FAQ",
+          "FAQ",
+        ),
+        this.fetchDirectusJSON(
+          "https://admin.topwr.solvro.pl/items/FAQ_Types_FAQ",
+          "FAQ Types FAQ",
+        ),
```
mini-bomba, line 64: it's a bit unclear what `FAQ Types FAQ` means.
consider using `FAQ pivot table` instead.

database/scrapers/organizations.ts
```diff
@@ -106,9 +102,14 @@ export default class OrganizationsScraper extends BaseScraperModule {
...
-      const logoKey = org.logo !== null ? await this.newAsset(org.logo) : null;
+      const logoKey =
+        org.logo !== null
+          ? await this.directusUploadFieldAndGetKey(org.logo)
+          : null;
       const coverKey =
-        org.cover !== null ? await this.newAsset(org.cover) : null;
+        org.cover !== null
+          ? await this.directusUploadFieldAndGetKey(org.cover)
+          : null;
```
mini-bomba, line 112: use `.addErrorContext` on these to mention which fields these files were from

mini-bomba (changes requested):
a bunch of error handling related minor issues

database/scrapers/faq.ts
Leadman5555: Oh, I thought those were some frequently asked questions about the types of frequently asked questions. As in, each FAQ has a category and you have other Q&A that explain what that category is. 🤣

mini-bomba (approved):
:fire: :shipit: once ci passes after the rebase

PR #204, opened 2025-07-08
feat(mobile config): day swap lookahead field

Leadman5555:


start/routes.ts
```diff
@@ -98,18 +98,20 @@ router
...
+      .group(() => {
+        router.get("/", [MobileConfigController, "index"]);
+        router
+          .patch("/bump", [MobileConfigController, "bump"])
+          .use(middleware.auth());
+        router
+          .patch("/update_lookahead/:new_value", [
+            MobileConfigController,
+            "updateCount",
+          ])
```
mini-bomba, line 111: hell nah
that's now how you make rest endpoints
since the mobile config singleton will likely get more random fields, just define a generic update method that can update any of the mobile config fields - you can even use the base controller class and just not use the other methods it generates

start/routes.ts
```diff
@@ -98,18 +98,20 @@ router
...
-          "bump",
-        ])
-        .use(middleware.auth());
-    });
+    router
+      .group(() => {
+        router.get("/", [MobileConfigController, "index"]);
+        router
+          .patch("/bump", [MobileConfigController, "bump"])
+          .use(middleware.auth());
```
mini-bomba, line 106: i know this is something the previous PR screwed up, but it appears that it slipped through the review process
using `PATCH` for this kind of endpoint (one that doesn't accept a request body with the edits to make) doesn't make sense. a standard `POST` would be more fitting.

database/migrations/1751969302203_convert_cache_ref_to_mobile_configs_table.ts
```diff
@@ -0,0 +1,24 @@
...
+export default class extends BaseSchema {
+  protected tableName = "mobile_configs";
+  protected oldTableName = "cache_reference_numbers";
+
+  async up() {
+    this.schema.raw(
+      `ALTER TABLE ${this.oldTableName} RENAME TO ${this.tableName}`,
+    );
+    this.schema.alterTable(this.tableName, (table) => {
+      table.integer("day_swap_lookahead").unsigned().notNullable().defaultTo(3);
```
mini-bomba, line 12: why 3?

database/migrations/1751969302203_convert_cache_ref_to_mobile_configs_table.ts
```diff
@@ -0,0 +1,24 @@
+import { BaseSchema } from "@adonisjs/lucid/schema";
+
+export default class extends BaseSchema {
+  protected tableName = "mobile_configs";
+  protected oldTableName = "cache_reference_numbers";
+
+  async up() {
+    this.schema.raw(
+      `ALTER TABLE ${this.oldTableName} RENAME TO ${this.tableName}`,
+    );
```
mini-bomba, line 10: use `this.schema.renameTable` instead (and make sure to NOT await the returned promise)

database/migrations/1751969302203_convert_cache_ref_to_mobile_configs_table.ts
Leadman5555: Well, three sounds reasonable - not too late, but not too early. Mobile will change it to whatever they want anyway.

app/controllers/mobile_config_controller.ts
mini-bomba, line 1: you didnt need to remove the custom index implementation, or override default methods that should not be used.

here's a tip: you can override the [`$configureRoutes` method](https://github.com/Solvro/backend-topwr/blob/main/app/controllers/base_controller.ts#L410-L416) to configure custom routes for a controller extending the base controller.
if you want to keep all existing routes, call `super` in the implementation, otherwise just implement the method from scratch.
in there all you need to do is call router methods to add new routes - these are executed in a router grouped, with an automated prefix set to `/api/v1/<controller name>/`

mini-bomba:
I just noticed that the base controller's `update` method likely won't work without the id in the url.
All else looks good, so I'll just push a quick commit onto this branch that implements singleton support into base controller and merge.

Issue #207, opened 2025-07-12
[TICKET] - add to departments to distinct departments vs branches


simon-the-shark:
we're gonna support wroclaw tech branches  (legnica, jelenia itp) and we need a way to distinct them in the departments view. Either `is_branch` boolean, or enum or whatever you want. 

**LOW PRIO**

PR #209, opened 2025-07-15
feat: department distinction field

Leadman5555:
For #207


database/scrapers/departments.ts
mini-bomba, line 1: `isBranch` isn't present on the directus resources

mini-bomba (changes requested):

database/scrapers/departments.ts
Leadman5555: I think so. The issue says "we are going to support", so I suppose they will. I'll ask on discord

Leadman5555: > I think so. The issue says "we are going to support", so I suppose they will. I'll ask on discord

They are not going to add it to directus, but everything from directus will not be a branch, so `isBranch` is false for everything in scrapers

mini-bomba (approved):
:fire: :shipit:

PR #210, opened 2025-07-15
fix: range search

Leadman5555:
Range search did somewhat work, but it had trouble parsing the request parameters. I changed the logic a bit and it works correctly now as long as the range search JSON is encoded properly - more about it in the updated README

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,19 +222,18 @@ Examples:
 - get all campuses whose names end with `a`:<br>
   `GET https://api.topwr.solro.pl/api/v1/campuses?name=%a`
 
-Any numeric or date-time field can be filtered by range. Pass the following JSON object as a query param: `{"from": "<value>", "to": "<value>"}`
+Any numeric or date-time field can be filtered by range. Pass the following query params to filter by range:
+
+- for lower bound of greater or equal to value: `<param>.from=<value>`
+- for upper bound of lesser or equal to value: `<param>.to=<value>`
+- for both lower and upper bound, pass both params separately (that is: `<param>.from=<value>&<param>.to=<value>`)
 
 You can pass both `from` lower bound and `to` upper bound or only one of them.
-When both are passed, if `from` > `to`, the request will result in a 400 Bad Request response.
-If `from` is equal to `to`, the filter works exactly like the single value filter (in other words: `departmentId={"from": "1", "to": "1"}` is equal to `departmentId=1`).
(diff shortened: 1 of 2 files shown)


app/scopes/search_helper.ts
```diff
@@ -101,15 +97,25 @@ export function handleSearchQuery<T extends LucidModel>(): QueryScope<
...
   });
 }
 
+function checkForRangeSearch(value: string): FromTo | undefined {
+  try {
+    return JSON.parse(value) as FromTo;
+  } catch {
+    return undefined;
+  }
+}
```
mini-bomba, line 117: this check is actually insufficient - remember that `JSON.parse` can parse more than just object, including numbers.
this check will actually treat any number as a range query, thus breaking search by exact value for numbers.

README.md
```diff
@@ -210,22 +210,31 @@ For example, to sort campuses by `name`, in ascending order, send the following
...
+When both are passed, if `from` > `to`, the request will result in a 400 Bad Request response.
+If `from` is equal to `to`, the filter works exactly like the single value filter (in other words: `departmentId={"from": "1", "to": "1"}` is equal to `departmentId=1`).
+
+**Remember to encode the JSON properly before sending**.
+If the JSON is not encoded, it will be treated as two separate query parameters (with respective values `"from":"<value>"`,`"to":"<value>"`),
+and the request will either fail or yield an unintended result.
+
+Example of a correctly encoded request:  
+JSON: `{ "from": "3", "to": "4" }`  
+Request: `GET https://api.topwr.solvro.pl/api/v1/fields_of_study?departmentId=%7B%22from%22%3A%223%22%2C%22to%22%3A%224%22%7D`
```
mini-bomba, line 237: while this query scheme is functional, i feel like it's a bit... ugly
i wonder if we could improve it... maybe if instead using json, we just accepted query params with `.from`/`.to` suffixes, such as `departmentId.from=3&departmentId.to=4`?
that's an optional thing tho, if you're up for it

mini-bomba (commented):
just a minor issue and a minor improvement proposal

README.md
Leadman5555: I thought of this as well - mainly because the need to encode the json makes making quick curl testing requests tedious. The current logic would need be changed a bit, but for the better, I suppose. And we'd get rid of the parsing caveats too. I'll change it 👍

Issue #145, opened 2025-04-06
feat: add banner for mobile app


czaja307:

mini-bomba:
none, although with base controller now adapted to singletons, this should be dead simple to complete.

PR #215, opened 2025-07-20
feat: banners

AlicjaBonar:
Related to #145
Adds migration, model and controller for banner


app/models/banner.ts
```diff
@@ -0,0 +1,50 @@
...
+  declare title: string;
+
+  @typedColumn({ type: "string", optional: false })
+  declare description: string;
+
+  @typedColumn({ type: "string", optional: true })
+  declare url: string | null;
+
+  @typedColumn({ type: "boolean", optional: false })
+  declare isEnabled: boolean;
```
czaja307, line 24: I don't thing we need to store isEnabled if we're gonna use `visibleFrom` and `visibleUntil`. I'd go with computed property instead: https://lucid.adonisjs.com/docs/serializing-models#computed-properties

@simon-the-shark what do you think?

mini-bomba: obviously if we have start/end times for visibility, the "is visible" property should be computed from those values
tho a setting such as "draft" that would force a banner to always be invisible, even if it's in the set range could be useful

app/models/banner.ts
```diff
@@ -0,0 +1,50 @@
...
+  declare id: number;
+
+  @typedColumn({ type: "string", optional: false })
+  declare title: string;
+
+  @typedColumn({ type: "string", optional: false })
+  declare description: string;
+
+  @typedColumn({ type: "string", optional: true })
+  declare url: string | null;
```
mini-bomba, line 21: this property could likely use a custom validator that would only allow urls

app/models/banner.ts
```diff
@@ -0,0 +1,50 @@
...
+  declare isEnabled: boolean;
+
+  @typedColumn({ type: "number", optional: true })
+  declare textColor: number | null;
+
+  @typedColumn({ type: "number", optional: true })
+  declare backgroundColor: number | null;
+
+  @typedColumn({ type: "number", optional: true })
+  declare titleColor: number | null;
```
mini-bomba, line 33: we should have some specific format these values should be in and it should be enforced by custom validators

app/models/banner.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  @typedColumn.dateTime({ optional: true })
+  declare visibleFrom: DateTime | null;
+
+  @typedColumn.dateTime({ optional: true })
+  declare visibleUntil: DateTime | null;
```
mini-bomba, line 39: would be nice to enforce `visibleFrom` < `visibleUntil` on the validator/controller level, but you'd likely need to write update hooks for the controller like we did for the event calendar controller

app/models/banner.ts
mini-bomba: discussed on weekly: we would like such `draft` field

mini-bomba:
I uncovered a minor bug in the typed model annotations for datetimes where the validator was generated incorrectly for optional fields, converting null values to undefined and making it impossible to reset a datetime field to null

PR #222, opened 2025-07-29
feat(ci): link reminder

### LINKED ISSUE #218: ci: issue link reminder
add a github actions workflow, triggered on pull request creation:
- probably wait a few seconds for github to register all initial PR actions or for the author to link an issue (maybe a minute or so in that case?)
- check whether the pull request has linked issues
- if it doesn't, post a comment nagging the author to link an issue to their PR

Leadman5555:
Closes #218 

The script can only detect linking via the PR body (just like I did in this PR two lines above). It is not able to detect manual linking from the sidebar as it is not exposed in the PR event timeline (with events such as comments, reopenings and such). I've checked if there were any alternative solutions online, but other similar scripts are also unable to detect manual linking (more information about this can be found in the readme of this repo https://github.com/nearform-actions/github-action-check-linked-issues ).


.github/workflows/link_reminder.yml
```diff
@@ -0,0 +1,33 @@
+name: Link reminder
+
+on:
+  pull_request:
+    types: [opened, reopened]
```
mini-bomba, line 5: you probably shouldn't run the workflow on PR reopens, as it would then post a duplicate comment

.github/workflows/link_reminder.yml
```diff
@@ -0,0 +1,33 @@
+name: Link reminder
+
+on:
+  pull_request:
```
mini-bomba, line 4: The workflow got triggered on this pull request, because you specified the `pull_request` event, which uses workflows from the branch being merged.
There's an issue with that event though - while it will mostly work in our case, as we all work on one repository (without forks), it will likely break if someone makes a PR from a fork (or if dependabot makes a PR, as it is treated as a fork PR) as workflow runs from forks do not get write permissions in the tokens, even if you specify such permissions.
It also likely won't trigger if someone makes a PR if they branched out before this PR is merged.

That's only a minor issue though in our case, but if you do decide to change the event type, you could merge this workflow into the `pr_automation.yml` file.

.github/workflows/link_reminder.yml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  check_issue_link:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: read
+      pull-requests: write
+    steps:
+      - name: Wait
+        run: sleep 5
```
mini-bomba, line 16: if you're only matching on the PR body provided by the event payload, then this wait is likely unnecessary (even though I did specify it in the task)

mini-bomba (commented):
:fire: job, just a few minor issues/suggestions

.github/workflows/link_reminder.yml
Leadman5555: I moved the logic to the `pr_automation` workflow

.github/workflows/pr_automation.yml
```diff
@@ -2,11 +2,11 @@ name: Pull request automation
 
 on:
   pull_request_target:
-    type: opened
+    types: opened
```
mini-bomba, line 5: lol

.github/workflows/pr_automation.yml
```diff
@@ -23,3 +23,20 @@ jobs:
...
+            const prNumber = context.payload.pull_request.number;
+            const prBody = context.payload.pull_request.body || "";
+            const issueLinkRegex = /(?:fix(?:e[sd])?|close(?:[sd])?|resolve(?:[sd])?):?\s+#\d+/i;
+            if(!issueLinkRegex.test(prBody)){
+              await github.rest.issues.createComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: prNumber,
+                body: "Looks like you did not link an issue to this PR. If this PR completes a task, consider linking it."
+              });
```
mini-bomba, line 41: I edited the comment content to be a bit friendlier

mini-bomba (approved):
:fire: :shipit:

PR #225, opened 2025-08-24
feat: permissions 

### LINKED ISSUE #1: chore/moderation system
think about the way that we want to implement users with custom roles (students, workers, organizations, science clubs etc.)

these roles should have access to unique contents about themselves, which they can add or edit and then we need to accept those changes 

typical moderation and censorship stuff

czaja307:
closes #1 

link to ACL lib: https://github.com/holoyan/adonisjs-permissions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,6 @@ jobs:
         ports:
           - 5432:5432
     env:
-      ADMIN_RELATIONS_KEY: ${{ secrets.ADMIN_RELATIONS_KEY }}
       DB_HOST: localhost
       DB_PORT: 5432
       DB_USER: postgres


app/controllers/base_controller.ts
```diff
@@ -128,6 +146,83 @@ export default abstract class BaseController<
...
+   */
+  protected async authorizeRecord(
+    http: HttpContext,
+    action: ControllerAction,
+    record: InstanceType<T>,
+  ): Promise<void> {
+    void http;
+    void action;
+    void record;
+  }
```
mini-bomba, line 224: authorization should ideally be done before interacting with the db, so it should ideally take an id, not a record

app/controllers/guide_article_drafts_controller.ts
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    const rawId: unknown = (params as unknown as { id?: string | number }).id;
+    let draftId: number | undefined;
+    if (typeof rawId === "string") {
+      const numeric = Number(rawId);
+      draftId = Number.isNaN(numeric) ? undefined : numeric;
+    } else if (typeof rawId === "number") {
+      draftId = rawId;
+    }
```
mini-bomba, line 212: you should use an autogenerated validator here (check how base controller implementations do this)

app/controllers/guide_article_drafts_controller.ts
```diff
@@ -0,0 +1,260 @@
...
+    if (draftId === undefined) {
+      const { NotFoundException } = await import("#exceptions/http_exceptions");
+      throw new NotFoundException();
+    }
```
mini-bomba, line 216: bad request, actually (tho the autogenerated validator would throw the correct exception here)

app/controllers/guide_article_drafts_controller.ts
```diff
@@ -0,0 +1,260 @@
...
+    const draft = await GuideArticleDraft.find(draftId);
+    if (draft === null) {
+      const { NotFoundException } = await import("#exceptions/http_exceptions");
+      throw new NotFoundException();
+    }
```
mini-bomba, line 221: use `.findOrFail` + maybe `.addErrorContext` - you're already authenticated & authorized, no need to hide information from us

app/controllers/guide_article_drafts_controller.ts
```diff
@@ -0,0 +1,260 @@
...
+    if (draft.originalArticleId !== null) {
+      const existingArticle = await GuideArticle.find(draft.originalArticleId);
+      if (existingArticle === null) {
+        const { NotFoundException } = await import(
+          "#exceptions/http_exceptions"
+        );
+        throw new NotFoundException(
+          `Original article with id ${draft.originalArticleId} not found`,
+        );
+      }
```
mini-bomba, line 242: `.findOrFail` + `.addErrorContext`

app/controllers/guide_article_drafts_controller.ts
```diff
@@ -0,0 +1,260 @@
...
+      if (draft.imageKey === null) {
+        const { BadRequestException } = await import(
+          "#exceptions/http_exceptions"
+        );
+        throw new BadRequestException("Cannot create article without image");
+      }
+      article = await GuideArticle.create(draftData);
+    }
+
+    await draft.delete();
```
mini-bomba, line 256: add error context here
also if this fails, we throw an error, the draft still exists but also the guide article gets updated...
is this the time to use a database transaction?

app/controllers/permissions_controller.ts
```diff
@@ -0,0 +1,226 @@
...
+    };
+    const resourceType = resource.type;
+    const resourceName = resource.name;
+    if (resourceType !== "class" && resourceType !== "model") {
+      return { success: false, error: "Invalid resource.type" };
+    }
+    if (!isResourceKey(resourceName)) {
+      return { success: false, error: "Invalid resource.name" };
+    }
+    const userId = parsedUserId;
```
mini-bomba, line 92: please use vine validators and throw errors

app/controllers/permissions_controller.ts
```diff
@@ -0,0 +1,226 @@
...
+
+    const targetUser = await User.find(userId);
+    if (targetUser === null) {
+      throw new NotFoundException("User not found");
+    }
```
mini-bomba, line 97: `.findOrFail` + `.addErrorContext`

app/controllers/permissions_controller.ts
```diff
@@ -0,0 +1,226 @@
...
+
+    const Model = resourceRegistry[resourceName];
+
+    // Try common revoke method names; if library exposes a different API, TS will guide fixes.
+    const manager = Acl.model(targetUser) as unknown as {
+      disallow?: (action: ActionSlug, target: unknown) => Promise<void>;
+      revoke?: (action: ActionSlug, target: unknown) => Promise<void>;
+      remove?: (action: ActionSlug, target: unknown) => Promise<void>;
+    };
```
mini-bomba, line 185: why do we even need these casts? isn't there a type available for this interface?

app/controllers/permissions_controller.ts
```diff
@@ -0,0 +1,226 @@
...
+
+    if (resourceType === "class") {
+      if (typeof manager.disallow === "function") {
+        await manager.disallow(action, Model);
+      } else if (typeof manager.revoke === "function") {
+        await manager.revoke(action, Model);
+      } else if (typeof manager.remove === "function") {
+        await manager.remove(action, Model);
+      } else {
+        return { success: false, error: "Revoke not supported by ACL API" };
```
mini-bomba, line 195: this should be a 500

app/controllers/permissions_controller.ts
mini-bomba, line 1: an endpoint to list a user's permissions could be useful

app/controllers/student_organization_drafts_controller.ts
mini-bomba, line 1: i wonder if you could figure out a way to make a generic draft controller, so that you don't have to duplicate code for each model that may have drafts

app/decorators/typed_model.ts
```diff
@@ -462,8 +462,11 @@ export function typedManyToMany<Related extends LucidModel>(
   model: () => Related,
   options: TypedManyToManyOptions<Related>,
 ) {
-  options.meta ??= { declaredColumnTypes: {} };
-  options.meta.declaredColumnTypes = Object.fromEntries(
+  const meta = (options.meta ?? { declaredColumnTypes: {} }) as {
+    declaredColumnTypes: Record<string, RelationColumnDef>;
+  };
+  options.meta = meta;
```
mini-bomba, line 468: should've done this instead
```ts
options.meta ??= { declaredColumnTypes: {} };
const meta = options.meta;
```

app/models/guide_article.ts
```diff
@@ -17,7 +22,15 @@ import { handleSortQuery } from "#scopes/sort_helper";
...
-export default class GuideArticle extends BaseModel {
+@MorphMap("guide_articles")
+export default class GuideArticle
+  extends BaseModel
+  implements AclModelInterface
+{
+  getModelId(): ModelIdType {
+    return this.id;
+  }
+
```
mini-bomba, line 33: solvronis: digs deep into adonis internals to figure out which column is the primary key
this library: yeah whatever, just give use a method, we're too lazy

app/models/guide_article_draft.ts
```diff
@@ -0,0 +1,68 @@
...
+  declare title: string;
+
+  @typedColumn({ type: "string" })
+  declare shortDesc: string;
+
+  @typedColumn({ type: "string" })
+  declare description: string;
+
+  @typedColumn({ foreignKeyOf: () => FileEntry, optional: true })
+  declare imageKey: string | null;
```
mini-bomba, line 41: why is the image key optional, while other fields are not nullable?

app/controllers/guide_article_drafts_controller.ts
```diff
@@ -0,0 +1,260 @@
...
+        throw new ForbiddenException();
+      }
+    }
+  }
+
+  // When linking to an existing article, require per-model permission on the original article
+  protected async storeHook(ctx: {
+    http: HttpContext;
+    request: PartialModel<typeof GuideArticleDraft>;
+  }) {
```
mini-bomba, line 131: what if multiple drafts are created for the same article?

app/permissions/permissions.ts
```diff
@@ -0,0 +1,30 @@
...
+  slug: "read",
+});
+
+export const destroy = await Acl.permission().create({
+  slug: "destroy",
+});
+
+export const approveDraft = await Acl.permission().create({
+  slug: "approve_draft",
+});
```
mini-bomba, line 22: this permission appears to be unused, as you require `solvro_admin` everywhere to approve drafts

database/migrations/1757361155456_create_student_organization_drafts_table.ts
```diff
@@ -0,0 +1,87 @@
...
+    this.schema.createTable(this.tableName, (table) => {
+      table.increments("id");
+      table.text("name").notNullable();
+      table.boolean("is_strategic").notNullable().defaultTo(false);
+
+      table.integer("department_id").unsigned().nullable();
+      table
+        .foreign("department_id")
+        .references("departments.id")
+        .onDelete("SET NULL");
```
mini-bomba, line 16: do we actually want to silently set department id to null if it's deleted?

database/migrations/1757361155457_create_guide_article_drafts_table.ts
```diff
@@ -0,0 +1,33 @@
...
+      table
+        .foreign("image_key")
+        .references("file_entries.id")
+        .onDelete("RESTRICT");
+
+      table.integer("original_article_id").unsigned().nullable();
+      table
+        .foreign("original_article_id")
+        .references("guide_articles.id")
+        .onDelete("SET NULL");
```
mini-bomba, line 23: tbh i'd either restrict or cascade here

app/controllers/guide_article_drafts_controller.ts
```diff
@@ -0,0 +1,286 @@
...
+   * - Different users to propose different changes to the same article
+   * - A user to create multiple draft versions before deciding which to submit
+   * - Parallel editing workflows
+   *
+   * The solvro_admin can approve/reject drafts individually, choosing which changes to apply.
+   */
+  protected async storeHook(ctx: {
+    http: HttpContext;
+    request: PartialModel<typeof GuideArticleDraft>;
+  }) {
```
mini-bomba: unaddressed concern from previous review: what if multiple drafts are created for the same article?

app/controllers/guide_article_drafts_controller.ts
```diff
@@ -0,0 +1,286 @@
...
+      params: { id: draftId },
+    } = (await request.validateUsing(this.pathIdValidator)) as {
+      params: { id: string | number };
+    };
+
+    // Use findOrFail with error context
+    const draft = await GuideArticleDraft.findOrFail(draftId).addErrorContext(
+      () =>
+        `GuideArticleDraft with id ${draftId} not found or user lacks permission`,
+    );
```
mini-bomba: the "user lacks permission" part seems just wrong
i'd get it if you would throw the exact same error on permission check failures, but you don't...

app/controllers/student_organization_drafts_controller.ts
```diff
@@ -0,0 +1,361 @@
...
+
+    // For non-admins, we cannot filter by pivot (no manual relation). Optional: keep unfiltered index,
+    // per-id endpoints are protected by authorizeById. If index must be filtered, add custom scope here
+    // using ACL tables. For now, leave index unrestricted after auth.
+
+    if (page === undefined && limit === undefined) {
+      return { data: await baseQuery };
+    }
+    return await baseQuery.paginate(page ?? 1, limit ?? 10);
+  }
```
mini-bomba: can you explain how this method differs from the base controller index method?

mini-bomba (changes requested):
i would've liked to have just merged this pr now (i've delayed this review for long enough already, and i'm not feeling good knowing that, sorry)
but honestly, i have a few actually major concerns and some of those aren't just one line fixes.

mini-bomba (changes requested):
I've read all code changes, and made my own modifications where I knew how ([see diff](https://github.com/Solvro/backend-topwr/compare/77f90e93..9dbd1ca))
I'm satisfied with the code so far, but I've encountered issues while testing the live behaviour:
- permissions don't seem to be automatically created - this likely needs a migration
  - i see that there's the /app/permissions/permissions.ts file, but it's not imported anywhere
  - also we may want to implicitly upgrade all current users to admins
  - this issue prevented me from testing any further
- theoretical minor issue: does the library delete row-level permissions when the corresponding row is deleted?

PR #226, opened 2025-08-25
feat: add branch fields to map objects and student organizations

### LINKED ISSUE #221: feat(models): add "branch" fields
We need to add `branch` fields to all map objects and student organizations, which will determine to which branch these objects belong to.

The mobile team suggests we use enums as the type: one for each branch, plus probably a `MAIN` value for stuff in wrocław? (or null?)

Anything that currently exists in the legacy CMS and the current "production" database belongs to the "main" branch - you can safely set all new fields to the "main" branch value, but make sure to remove the default value after migrating.

[discussion link for reference](https://discord.com/channels/687360174377533442/1305494727679938613/1399484099617624176)

AlicjaBonar:
- Adds 'branch' enum type field to all map object tables and the student_organizations table
- Sets default value to 'main' temporarily for existing records
- Removes default value after migration as was requested

Closes #221


database/migrations/1756142084606_remove_default_main_branch_from_map_objects_and_student_organizations_table.ts
```diff
@@ -0,0 +1,31 @@
...
+    }
+  }
+
+  async down() {
+    for (const tableName of this.tablesToChange) {
+      await this.schema.raw(
+        `ALTER TABLE "${tableName}" ALTER COLUMN "branch" SET DEFAULT 'main'`,
+      );
+    }
+  }
```
mini-bomba, line 30: do not await these - awaiting causes changes during dry runs, triggering check failures

app/enums/branch.ts
```diff
@@ -0,0 +1,27 @@
...
+  Walbrzych = "walbrzych",
+  Legnica = "legnica",
+}
+
+export const branchEnumValues = {
+  availableValues: [
+    {
+      value: Branch.Main,
+      label: "Main branch",
+    },
```
mini-bomba, line 13: "Main campus" would probably be a better description here.
this whole `branchEnumValues` constant likely won't be used for much longer anyway, as we're trying to get rid of the adminjs panel

database/migrations/1756142084606_remove_default_main_branch_from_map_objects_and_student_organizations_table.ts
mini-bomba, line 1: minor: you should be able to merge this migration with the previous one that adds the columns (assuming you don't await the query, adonis moment)

mini-bomba (commented):
very nice, just a few comments :shipit:

PR #232, opened 2025-09-16
feat(auth): change password endpoint

### LINKED ISSUE #231: feat: change password endpoint separate from the "i forgor 💀" form
basically allow users to update passwords using their already currently active session

karolwolek:
This closes #231.

As in the issue description, now the logged in user can reset it's password after providing currently used password, new one and the confirmation (correct me if what i did is okay for the backend endpoint or it should be only handled by the frontend.)
The main logic happens in `changePasswordValidator`. It fails if:
- current password is wrong
- new password does not meet criteria (8 chars, one upper, one lower case letter)
- new password is the same as the old one
- the confirm does not match

I moved the reset/change password logic to the auth controller as it is a more logical place for me with all the `login`, `logout` etc flows that are there. Also there is no need for `/admin` endpoint since we as backend team do not implement admin panel anymore so i also changed the route to the standard one: `/api/v1/auth`

sidenote:
while doing this pr, i was reminded by the stale issue in our code base that *forgot password email* is not working because it's providing a put request url. We need to cooperate somehow with frontend to put their form page url there

### DIFF
--- a/start/routes.ts
+++ b/start/routes.ts
@@ -82,7 +82,7 @@ router
             router.put("/:token", [AuthController, "resetPassword"]);
           })
           .prefix("/reset_password");
-        router.put("/change_password", [AuthController, "changePassword"]);
+        router.post("/change_password", [AuthController, "changePassword"]);
       })
       .use(middleware.sensitive())
       .prefix("/auth");


app/controllers/auth_controller.ts
```diff
@@ -31,4 +41,66 @@ export default class AuthController {
...
+   */
+  async changePassword({ request, auth }: HttpContext) {
+    if (!auth.isAuthenticated) {
+      await auth.authenticate();
+    }
+    const user = auth.getUserOrFail();
+
+    const form = await request.validateUsing(changePasswordValidator(user));
+    await user.updatePassword(form.newPassword);
+  }
```
mini-bomba, line 56: this function should probably return a simple success message

mini-bomba (commented):
very nice, just a few minor issues :shipit: 
(mostly about reimplementing existing vine functions)

PR #259, opened 2025-10-15
feat: serverside image resizing

### LINKED ISSUE #251: feat: serverside image resizing
full resolution images take up too much bandwidth and load too slowly - we should also serve downsized images

Leadman5555:
Closes #251 

Key word: "Photo-like" - a file is photo-like if its extension matches a set of predefined extensions (currently `const PHOTO_LIKE_EXT = ["png", "jpg", "jpeg", "webp"];` . Only photo-like files have miniatures.

#### Env variables:

- `MINIATURE_MAX_HEIGHT_PX=300 // max height of miniature, smaller files do not get resized. The width auto-scales so that images are not stretched`
- `MINIATURE_MAX_PROCESSING_TIME_S=5 // max time in seconds that resizing of an image can take before throwing. Although during testing conversions were basically instant.`

#### Miniatures serving:

- support for miniature files (available under `/uploads/miniatures` with the same key as the original file under the `/uploads`)
- every new uploaded photo-like file has its miniature automatically computed during upload. Uploading interface was not changed
- automatic computation of miniatures also works with scrappers
- removing a photo-like file also automatically removes its miniature if it exists

#### Miniature management:

- command `generate_miniatures` that will generate all missing miniatures for every photo-like file saved in the database (to prevent generating for orphaned files). Passing `--all` parameter regenerates all miniatures, not only the missing ones.
- command `files_cleanup` has an additional stage added that removes all orphaned miniatures

### Refactors:
- the `--force` parameter workaround was made more reusable
- simplified the structure of the `files_cleanup` command a bit

------

I've tested the conversion and scripts locally.

Before resize (available under `/uploads/0a9b164b-87b5-4005-b8f1-3ba8ad1f7436.jpg`):
image

After resize (available under `/uploads/miniatures/0a9b164b-87b5-4005-b8f1-3ba8ad1f7436.jpg`):
image

### DIFF
--- a/app/exceptions/file_service_errors.ts
+++ b/app/exceptions/file_service_errors.ts
@@ -9,7 +9,7 @@ class FileServiceFSError extends FileServiceError {
     super(`${driveType}: ${message}`, {
       code: "E_FILE_SYSTEM_ERROR",
       cause,
-      extraResponseFields: {
+      extraErrorFields: {
         driveType,
       },
     });
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "@adonisjs/mail": "^9.2.2",
         "@adonisjs/session": "^7.5.0",
         "@adonisjs/static": "^1.1.1",
-        "@solvro/error-handling": "^1.0.0",
+        "@solvro/error-handling": "^1.1.0",
         "@solvro/solvronis-metrics": "^1.0.0",
         "@solvro/utils": "^1.2.0",
         "@vinejs/vine": "^2.1.0",
@@ -4221,10 +4221,13 @@
       "license": "MIT"
     },
     "node_modules/@solvro/error-handling": {
-      "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/@solvro/error-handling/-/error-handling-1.0.0.tgz",
-      "integrity": "sha512-1GNDMrM6Og95qBWkCoiCOzJnACxOw+8Ed6ueZ00v+tUq+Chf0qe4wnJv5O82PqxY8n2otPHGEe5kF+QYUtSsyw==",
-      "license": "MPL-2.0"
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/@solvro/error-handling/-/error-handling-1.1.0.tgz",
+      "integrity": "sha512-edxInRyBLlSBEeDgJMIquB5upXRhHzPBPU9p08nmUnR+ebnYZhM7kgEcmjS6+0PHloQLgHVw5Y5/y+jvd0ypQg==",
+      "license": "MPL-2.0",
+      "dependencies": {
+        "@solvro/utils": "^1.4.0"
+      }
     },
     "node_modules/@solvro/solvronis-metrics": {
       "version": "1.0.0",
@@ -4237,9 +4240,9 @@
       }
     },
     "node_modules/@solvro/utils": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/@solvro/utils/-/utils-1.2.0.tgz",
-      "integrity": "sha512-B/me7DOnqlFbS5iuEeM9+H8Vu5x8b60M4D0tpbcs8+3lUBPtGMZTz8E/hcD9q/yCnApY6yK9f4l4lA0H8nHdiA==",
+      "version": "1.4.0",
+      "resolved": "https://registry.npmjs.org/@solvro/utils/-/utils-1.4.0.tgz",
+      "integrity": "sha512-Xdhl9PMVdHYsu4CXTS+9tF0bepMdCRPW8ylmZvD5Slmjq6O87b3D5MI7V+iSAavHaB3zMaOBpQK2HwJmKG1yPg==",
       "license": "MPL-2.0"
     },
     "node_modules/@stylistic/eslint-plugin": {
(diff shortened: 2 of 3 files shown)


app/exceptions/file_service_errors.ts
```diff
@@ -1,12 +1,17 @@
...
 
 class FileServiceFSError extends FileServiceError {
-  constructor(message: string, cause: unknown) {
+  constructor(message: string, cause: unknown, driveType: DriveType) {
     super(message, {
       code: "E_FILE_SYSTEM_ERROR",
       cause,
+      extraResponseFields: {
+        driveType,
+      },
```
mini-bomba: i guess i should add `extraErrorFields` to the error handling library?
cause this makes more sense as a property of the error, and not as a field on the response?

commands/base_command_extended.ts
mini-bomba: this is actually quite smart

commands/base_command_extended.ts
```diff
@@ -0,0 +1,64 @@
...
+ */
+export interface SupportsListing {
+  list(): string;
+}
+
+export default abstract class BaseCommandExtended extends BaseCommand {
+  // Apparently, Adonis tries to serialize everything derived from BaseCommand, even if it is abstract.
+  // Thus, the name property is required for it to compile.
+  static commandName = "abstract_command";
+  static description = "Calling this command does nothing";
```
mini-bomba: i wonder if removing the default export (and making it named instead) would solve this issue

commands/files_cleanup.ts
```diff
@@ -300,11 +346,17 @@ export default class CleanupFiles extends BaseCommand {
     });
   }
 
+  private removeFromDisk(entry: LocalFileEntry) {
+    const key = `${entry.uuid}${entry.ext}`;
+    // Miniature first in case of a failure as it can be recomputed
+    fs.rmSync(path.join(this.miniaturesStoragePath, key));
+    fs.rmSync(path.join(this.storagePath, key));
+  }
```
mini-bomba: does `fs.rmSync` throw if the file does not exist?
if so, this might make it impossible to remove non-image files or images with no computed miniatures

commands/files_cleanup.ts
```diff
@@ -51,25 +55,18 @@ export default class CleanupFiles extends BaseCommand {
   };
 
   @flags.boolean({
-    description: "Run all cleanup stages without asking",
+    description:
+      "Run all cleanup stages without asking. Use default values for all prompts",
   })
   declare force: boolean;
```
mini-bomba: minor nitpick about the description: setting force does technically change the "default" values for prompts for deletion from false to true

mini-bomba (commented):
Really nice work :fire: 
this is just the initial round of review comments, with mostly nitpicks/suggestions (and one possibly "minor" issue)
i didn't review the use of the `sharp` library in-depth yet, and this is a to-do for the next review

app/exceptions/file_service_errors.ts
Leadman5555: I'll make a PR there then

app/services/files_service.ts
```diff
@@ -63,17 +86,23 @@ export default class FilesService {
     extname: string | undefined = undefined,
   ): Promise<FileEntry> {
     const fileEntry = FileEntry.createNew(extname);
+    if (fileEntry.isPhoto()) {
+      // To compute the miniature, we need to consume the stream. We also need its content in two places (miniature and original)
+      const rawData = await stream.toArray();
+      const data = Buffer.concat(rawData);
+      return this.uploadFromMemoryInternal(data, fileEntry);
+    }
```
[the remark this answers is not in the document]
Leadman5555: I don't think it will. Calling the `FileEntry.createNew(extname)` creates an instance, but it is persisted in the transaction afterwards. If the stream throws, there will be a transaction rollback - that's the very reason I used transactions when implementing the FileService a while back.

commands/files_cleanup.ts
Leadman5555: That's a good spot - I'll add the `force` option
image

config/drive.ts
```diff
@@ -1,31 +1,50 @@
...
-  fs.mkdirSync(storagePath, { recursive: true });
+export function ensureStorageDirsExist() {
+  if (!fs.existsSync(MINIATURES_STORAGE_PATH)) {
+    fs.mkdirSync(MINIATURES_STORAGE_PATH, { recursive: true });
+  }
 }
 
+ensureStorageDirsExist();
+
+export const MAIN_DRIVE = env.get("DRIVE_DISK");
```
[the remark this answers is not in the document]
Leadman5555: Sure - I also found it a bit weird since the type of the var is defined as "fs" and not as a string

commands/base_command_extended.ts
mini-bomba: yeah no, this is an error thrown by adonis 😔

app/services/files_service.ts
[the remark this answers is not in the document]
mini-bomba: oh right, i forgot `.createNew` doesn't actually create the file entry in the database until it is saved

mini-bomba (approved):
usage of the library looks fine, which I think was the only thing I left for myself to review now.
gonna assume i covered everything else in my previous review
:fire: :shipit:

PR #261, opened 2025-10-29
feat: push notifications

### LINKED ISSUE #253: feat: push notifications
required parameters for the request:
- notif title
- notif body
- extra json data (deeplinks?)

all pushed via firebase

Leadman5555:
Closes #253 

#### Refactors

- small `BaseController` refactor - behaviour not changed, just extracted some repeated things to methods to use in inheritors

#### New things

- route configuration for `BaseController`: allows to skip chosen routes from the base CRUD
- firebase service key file (same as in sks repo - needed as credentials)
- `FirebaseTopic` entity and table to host topics and track their state
- `FirebaseTopicsController`:
  - hooked up to the `BaseController` with index, destroy routes disabled (prefixed with `/topic`)
  - `/topic?deactivatedSince=<timestamp>`: replaces index endpoint, fetches the list of all active topics and all topics that have been deactived since the given timestamp (good for performance, if the mobile only needs to fetch changes since last fetch)
 - state update handled via base `update` endpoint with extra hooks to ensure the state is valid
  - `/broadcast`: endpoint for broadcasting a notification to given topic(s) (if multiple topics are passed, the message will be sent to all of them). Will fail if any of the topics is deactivated/non-existent (invalid topics in error response). Body:
  ```json
  {
    "notification": {
      "title": "ok",
      "body": "okok",
      "data": {} //optional
    },
    "topics": ["test1","test2"] 
 }
 ```


app/controllers/base_controller.ts
```diff
@@ -106,6 +106,20 @@ export type DeleteHookContext<T extends LucidModel> = Omit<
...
+export interface RouteConfigurationOptions {
+  skipRoutes: {
+    index?: boolean;
+    store?: boolean;
+    show?: boolean;
+    destroy?: boolean;
+    update?: boolean;
+  };
+}
+
```
mini-bomba: that's a smart change

app/models/firebase_topic.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  /**
+   * Use this method to manipulate the topic state instead of manually field updates
+   * Note: This method does not change the database state
+   */
+  public activateTopic() {
+    this.isActive = true;
+    this.deactivatedAt = null;
+  }
```
mini-bomba: you probably could use an adonis model update hook for these?
this way you wouldn't need the dedicated activate topic endpoints

app/models/firebase_topic.ts
```diff
@@ -0,0 +1,102 @@
...
+  public static async getTopicState(
+    deactivatedSince: DateTime,
+  ): Promise<TopicState> {
+    const result = await db.rawQuery<{
+      rows: TopicState[];
+    }>("SELECT * FROM get_fb_topic_state(?)", [deactivatedSince.toSQL()], {
+      mode: "read",
+    });
+    return result.rows[0];
+  }
```
mini-bomba: this function & type name could use better naming - "topic state" kinda implies the current state of just one topic?

app/controllers/firebase_topics_controller.ts
```diff
@@ -0,0 +1,138 @@
...
+      .patch("/topic/toggle/:id", [controller, "toggleTopicState"])
+      .as("toggle");
+    router.get("/topic/state", [controller, "indexTopicState"]).as("state");
+    router
+      .post("/notification/broadcast", [
+        controller,
+        "broadcastPushNotification",
+      ])
+      .as("broadcast");
+  }
```
mini-bomba: the base controller methods here will not have `/topic` prefixes, while other methods will, which may be confusing.
i'd rename this controller to just `firebase_controller.ts`, and wrap the call to the base controller's configure route method in `router.group` with the `/topic` prefix.

mini-bomba (commented):
minor issues (and some praise)

mini-bomba (approved):
imma just trust my past self
:fire: :shipit:

app/controllers/base_controller.ts
```diff
@@ -489,15 +503,27 @@ export default abstract class BaseController<
...
+      }
+      if (!(configurationOptions?.skipRoutes.show === true)) {
+        router.get("/:id", [controller, "show"]).as("show");
+      }
+      if (!(configurationOptions?.skipRoutes.destroy === true)) {
+        router.delete("/:id", [controller, "destroy"]).as("destroy");
+      }
+      if (!(configurationOptions?.skipRoutes.update === true)) {
+        router.patch("/:id", [controller, "update"]).as("update");
+      }
```
kguzek: u was idiomatycznie jest tak zamiast np

```ts
if (configurationOptions?.skipRoutes.destroy !== true) {
  // ...
```
? 🤔

PR #267, opened 2025-11-11
feat: push history

### LINKED ISSUE #266: feat: toPWr Main - PUSH history
Sub-issue of #263 for the ToPWr main repo

Leadman5555:
Closes #266 

#### New things

- Notifications controller with one main index endpoint (POST because it is with body) - described in more detail in the readme docs
- Push notifications are automatically saved after attempting to send them with information if sending was successful
- basic JSON support for typedColumn

### DIFF
--- a/app/decorators/typed_model.ts
+++ b/app/decorators/typed_model.ts
@@ -243,10 +243,6 @@ function decoratorTypeToValidatorAndType(
         break;
       }
     }
-    return {
-      declaredType,
-      validator,
-    };
   } else {
     validator = vine.enum(Object.values(type));
     declaredType = "enum";


kguzek:
> I'll push a fix to the docs in a second. As for why it's not a GET, well, I think it's cleaner this way if you may have multiple excludes and includes - the query string is not super long. Though if you prefer the param version, I can change it - let me know about it in that case.

I would definitely prefer the query params version, e.g.
`GET /api/v1/firebase/notifications?include=topic1,topic2,topic3&exclude=topic4,topic5`

It would also improve performance because Next.js/the browser does not cache POST requests

generally frontend web dev tools are built around the assumption that GET requests do not modify anything, so they are safe to cache/refetch/revalidate if needed, and POST requests are only for mutations, so they should never be repeated or refetched, e.g. on receiving an error response

Thanks

Leadman5555:
Alright, I updated the index endpoint, you can now use the requested format: `GET /api/v1/firebase/notifications?include=topic1,topic2,topic3&exclude=topic4,topic5`, with one topic or topics separated by commas

app/models/push_notification_entry.ts
```diff
@@ -0,0 +1,82 @@
...
+  declare id: number;
+
+  @typedColumn({ type: "string" })
+  declare title: string;
+
+  @typedColumn({ type: "string" })
+  declare body: string;
+
+  @typedColumn({ type: "json", optional: true })
+  declare data: Record<string, string> | null;
```
mini-bomba: honestly, i'd just fake the type here to be something like string (but leave a comment about the type being fake)
manually creating this model should be blocked anyway
and declaring a `json` type in our typing framework feels weird - we likely won't use it anywhere else anyway

mini-bomba (commented):
almost an approve
just a few minor issues

PR #270, opened 2025-11-28
feat: DAS structure

### LINKED ISSUE #265: DAS Feature
Image

Leadman5555:
Closes #265 

Overall, has all the requested entities (details in the linked issue), as well as base controller endpoints for basic crud.  As for business logic, there is little to none involved.

### DIFF
--- a/app/controllers/v1/drafts.ts
+++ b/app/controllers/v1/drafts.ts
@@ -177,6 +177,7 @@ export abstract class GenericDraftController<
       case "update":
       case "destroy":
       case "relationIndex":
+      case "oneToOneRelationStore":
       case "oneToManyRelationStore":
       case "manyToManyRelationAttach":
       case "manyToManyRelationDetach":
@@ -238,6 +239,7 @@ export abstract class GenericDraftController<
       update: "update",
       destroy: "destroy",
       relationIndex: "read",
+      oneToOneRelationStore: "update",
       oneToManyRelationStore: "update",
       manyToManyRelationAttach: "update",
       manyToManyRelationDetach: "update",


app/models/das_stand.ts
```diff
@@ -0,0 +1,80 @@
...
+  declare dasId: number;
+
+  @belongsTo(() => Das, {
+    localKey: "id",
+    foreignKey: "dasId",
+  })
+  declare das: BelongsTo<typeof Das>;
+
+  @typedColumn({ type: "string" })
+  declare name: string;
```
mini-bomba: you declared `name` as a 127-char string in the database: you've been adding custom validators to enforce lengths everywhere else, so you might as well add it here

database/migrations/1764083040864_create_das_table.ts
mini-bomba: missing `updated_at`? (just for consistency with other models)

database/migrations/1764083217368_create_das_stands_table.ts
```diff
@@ -0,0 +1,40 @@
...
+        .inTable("student_organizations")
+        .nullable()
+        .onDelete("SET NULL");
+      table.text("description").nullable();
+      table
+        .uuid("logo_key")
+        .references("id")
+        .inTable("file_entries")
+        .nullable()
+        .onDelete("SET NULL");
```
mini-bomba: this relation should use the `RESTRICT` action

mini-bomba (changes requested):
our beloved minor issues (that collectively end up requiring actual changes)

app/models/das_timetable.ts
```diff
@@ -0,0 +1,42 @@
...
+import { handleSearchQuery } from "#scopes/search_helper";
+import { handleSortQuery } from "#scopes/sort_helper";
+
+export default class DasTimetable extends BaseModel {
+  @typedColumn({
+    isPrimary: true,
+    foreignKeyOf: () => Das,
+    autoGenerated: false,
+  })
+  declare id: number; // Id Mapping
```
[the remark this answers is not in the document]
Leadman5555: I think having it here is a good idea. The time slots shouldn't be linked directly to the event. Also, we might want to add more fields to the timetable. It's not a proxy but rather an association class, although mostly empty right now.

app/controllers/v1/das.ts
```diff
@@ -0,0 +1,16 @@
...
+  protected readonly queryRelations = [
+    "maps",
+    "maps.content",
+    "links",
+    "stands",
+    "stands.logo",
+  ];
+  protected readonly crudRelations = ["maps", "links", "stands"];
+  protected readonly model = Das;
+}
```
mini-bomba: these relations arrays are missing the 1:1 timetable relation, but 1:1 relations aren't supported here
if you don't mind diving into my spaghetti code, support for this should be added

app/models/das_stand.ts
```diff
@@ -0,0 +1,79 @@
...
+    optional: true,
+  })
+  declare description: string | null;
+
+  @typedColumn({
+    isPrimary: true,
+    foreignKeyOf: () => StudentOrganization,
+    optional: true,
+  })
+  declare studentOrganizationId: number | null;
```
mini-bomba: why is this set as a primary key?
also setting multiple keys as primary will likely confuse adonis

app/models/das_stand.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export default class DasStand extends BaseModel {
+  @typedColumn({ isPrimary: true, type: "integer", autoGenerated: true })
+  declare id: number;
+
+  @typedColumn({ type: "string", validator: vine.string().maxLength(7) })
+  declare number: string; // Knowing PWr, stand number could not be a number
+
+  @typedColumn({ isPrimary: true, foreignKeyOf: () => Das })
+  declare dasId: number;
```
mini-bomba: another dubious primary key declaration

PR #274, opened 2026-02-14
feat(event calendar): hide events logic

Leadman5555:
One new endpoint:
- PATCH `/event_calendar/hide/{googleCalId}?hide={booleanShouldHide}`
Will either hide an event or remove the hidden status depending on {booleanShouldHide} value. Since we might want to hide an event that is not yet scraped but will be, this does not rely on the events currently in the database.

One endpoint modified (optional param added)
- GET `/event_calendar?[showHidden={booleanShouldShowHidden}]`
If the parameter is not passed or false, will return only non-hidden events.
If parameter is passed and true, will return all events.

I also changed the type of googleCalId from text to varchar


app/controllers/v1/event_calendar.ts
mini-bomba: it would probably be nice to have an option to list all hidden event IDs as well, so that they can be unhidden easily

app/controllers/v1/event_calendar.ts
```diff
@@ -51,4 +81,52 @@ export default class EventCalendarController extends BaseController<
...
+    }
+    const {
+      hide,
+      params: { gCalId },
+    } = await request.validateUsing(hideToggleValidator);
+    const isAlreadyHidden =
+      (await db
+        .from("hidden_events")
+        .where("google_cal_id", gCalId)
+        .first()) !== null;
```
mini-bomba: tbh a "beg for forgiveness" approach is better here - instead of checking whether an entry exists, try to insert/delete it and check whether that operation succeeded
that will mean catching the primary key constraint violation error for inserts, and checking the number of rows affected for deletions

PR #279, opened 2026-02-23
feat(permissions): add cleanup command for orphaned permissions

czaja307:


app/utils/permissions.ts
```diff
@@ -34,3 +38,90 @@ const thinModelHandlers: ProxyHandler<ThinModel> = {
...
+    string,
+    unknown
+  >[];
+  const existingIds = existingRows.map((row) => row[primaryKey] as number);
+
+  // Base query: permissions scoped to this entity type at instance level
+  let query = db
+    .from("permissions")
+    .where("entity_type", entityType)
+    .whereNotNull("entity_id");
```
mini-bomba: is subquery type-casting an actual issue?
i'd imagine they'd be much more efficient if writing them wasn't an issue

PR #290, opened 2026-03-17
feat(rate limiting): implement upload rate limiter for non-admin users

### LINKED ISSUE #272: feat: upload ratelimiting for non-admins
to be implemented after #225 is merged
this probably shouldn't be too strict, but should prevent an untrusted user from filling up our disk

czaja307:
closes #272


mini-bomba (approved):
🔥 :shipit:

PR #292, opened 2026-03-21
refactor: base controller split

mini-bomba:
this pr:
- splits the BaseController into two:
  - the current "BaseContoller" was renamed to AutoCrudController
  - the self-routing method and superuser check helpers were moved to the new BaseController
    -  all controllers must now extend this BaseController
- route setup & controller import functions were moved to `utils/controllers.ts`
- each AutoCrudController now gets an autogenerated unit test that calls its self-validate method

### DIFF
--- a/app/controllers/auto_crud_controller.ts
+++ b/app/controllers/auto_crud_controller.ts
@@ -1009,7 +1009,7 @@ export default abstract class AutoCrudController<
     const mainInstance = await this.getFirstOrFail(id);
     await this.authorizeRecord(
       { request, route, auth } as unknown as HttpContext,
-      "oneToManyRelationStore",
+      "oneToOneRelationStore",
       mainInstance,
     );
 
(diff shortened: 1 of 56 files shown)


app/controllers/auto_crud_controller.ts
```diff
@@ -0,0 +1,1272 @@
...
+      relationName,
+    });
+    const toStore = (await request.validateUsing(
+      this.relatedStoreValidator(relationName),
+    )) as Partial<ModelAttributes<LucidRow>>;
+
+    const mainInstance = await this.getFirstOrFail(id);
+    await this.authorizeRecord(
+      { request, route, auth } as unknown as HttpContext,
+      "oneToManyRelationStore",
```
czaja307: shouldn't this one be `oneToOneRelationStore`?

mini-bomba: true, seems like i forgot when i was adding the new route

czaja307 (approved):
:fire: :shipit:

PR #298, opened 2026-04-30
fix(das): make timetable entry end time nullable

### LINKED ISSUE #293: fix(das): make timetable entry end time nullable

krzys13:
#293 
Make DasTimetableEntry.endTime nullable and update validation so null values are allowed for timetable entries that only have a start time.

Fix validator also and avoid errors when the end time is not known.

### DIFF
--- a/app/models/das_timetable_entry.ts
+++ b/app/models/das_timetable_entry.ts
@@ -24,9 +24,9 @@ export default class DasTimetableEntry extends BaseModel {
   declare startTime: DateTime;
 
   @typedColumn.dateTime({
-    validator: vine.luxonDateTime().afterField("startsAt"),
+    validator: vine.luxonDateTime().afterField("startTime").optional(),
   })
-  declare endTime: DateTime;
+  declare endTime: DateTime | null;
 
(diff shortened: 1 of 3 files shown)


mini-bomba (changes requested):
just editing the backend validations is not enough, you need to alter the db schema as well, otherwise you'll run into non-null constraint violations

database/migrations/1764088127037_create_das_timetable_entries_table.ts
mini-bomba, line 1: do not edit existing migrations, make a new one instead
run `node ace make:migration --alter` to get a fresh migration template

mini-bomba (changes requested):

database/migrations/1764088127037_create_das_timetable_entries_table.ts
krzys13: done

mini-bomba (approved):
i reverted an unintended change due to the merge commits and/or rebase
and made the migration use setNullable/dropNullable instead of changing the entire column type

🔥 :shipit:

PR #302, opened 2026-05-15
feat: user management api

### LINKED ISSUE #275: feat: user management api
we don't appear to have one at the moment???
should support user creation/modification/deletion
also need to add role assignment methods to the permissions controller
most actions should be solvro admin only, except for users modifying themselves (if it makes sense for them to modify themselves)

lolakk05:
I tested a few functionalities: 
- solvro_admin can change other users permissions, roles
- user cant change their own roles/permissions.


lolakk05:
aha i tak chyba źle podpiałem issue

mini-bomba:
the workflow that reminds you to link issues only looks at your pull request's description, so it only checks for issues linked with keywords, not manually
iirc github doesn't make it that easy to figure out which issues are linked manually (well they don't make it easy to check for keyword-linked issues either, but at least they make the list of valid keywords public)

mini-bomba:
also you don't need to manually request reviews on our repos, github automatically requests reviews from appropriate users/teams based on the codeowners file we've created

app/controllers/v1/permissions.ts
```diff
@@ -57,6 +64,8 @@ export default class PermissionsController extends BaseController {
     router.post("/allow", [controller, "allow"]).as("allow");
     router.post("/revoke", [controller, "revoke"]).as("revoke");
     router.get("/list", [controller, "listUserPermissions"]).as("list");
+    router.post("roles/assign", [controller, "assignRoles"]).as("assignRoles");
+    router.post("roles/revoke", [controller, "revokeRoles"]).as("revokeRoles");
```
mini-bomba: leading slashes missing in paths

app/controllers/v1/users.ts
mini-bomba, line 1: well i told you explicitly to make a custom controller, and we sure as hell don't want rando unauthenticated actors enumerating our users (while exposing their full name and email)

app/models/user.ts
```diff
@@ -61,6 +71,24 @@ export default class User
...
+  })
+  declare userRoles: ManyToMany<typeof Role>;
+
+  @manyToMany(() => Permission, {
+    pivotTable: "model_permission",
+    pivotForeignKey: "model_id",
+    pivotRelatedForeignKey: "permission_id",
+  })
+  declare userPermissions: ManyToMany<typeof Permission>;
+
```
mini-bomba: wat

mini-bomba (changes requested):
yeah no, this aint finished at all

app/controllers/v1/permissions.ts
lolakk05: done

app/controllers/v1/users.ts
lolakk05: done i think

app/controllers/v1/users.ts
```diff
@@ -0,0 +1,140 @@
...
+    if (!(await this.isSuperUser(auth)) && auth.user?.id !== userId) {
+      throw new ForbiddenException();
+    }
+  }
+
+  async findAll({ request, auth }: HttpContext) {
+    await this.requireSuperUser(auth);
+
+    const page = request.input("page", 1) as number;
+    const limit = request.input("limit", 10) as number;
```
mini-bomba: do not use input/param, validate everything (applies to the entire file)

app/controllers/v1/users.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    const users = await User.query()
+      .select("id", "fullName", "email")
+      .paginate(page, limit);
+
+    return { data: users };
+  }
+
+  async findOne({ request, auth }: HttpContext) {
+    await this.requireSuperUserOrSelf(auth, parseInt(request.param("id")));
```
mini-bomba: i think it would be more reasonable to have an endpoint such as `/users/me` rather than allowing the user to get "get" themselves using the standard endpoint if they know their own id
also we might already have such an endpoint, in the auth controller...

app/controllers/v1/users.ts
```diff
@@ -0,0 +1,140 @@
...
+      .select("id", "fullName", "email")
+      .where("id", userId)
+      .firstOrFail()
+      .addErrorContext(() => `User with id ${userId} not found`);
+
+    return { data: targetUser };
+  }
+
+  async delete({ request, auth }: HttpContext) {
+    await this.requireSuperUserOrSelf(auth, parseInt(request.param("id")));
```
mini-bomba: you should not be able to delete yourself, accounts are provisioned by solvro admins

app/controllers/v1/users.ts
```diff
@@ -0,0 +1,140 @@
...
+      () => `User with id ${userId} not found`,
+    );
+
+    await targetUser.delete();
+
+    return { success: true };
+  }
+
+  async update({ request, auth }: HttpContext) {
+    await this.requireSuperUserOrSelf(auth, parseInt(request.param("id")));
```
mini-bomba: and also be careful with what the user can update themselves
tbh i think they should only be able to edit their own password, and we likely have such a method in the auth controller already

app/models/user.ts
mini-bomba, line 1: not needed

mini-bomba (changes requested):
only general review pass for now, focusing on the source of your lint issues and possible api design problems

app/controllers/v1/permissions.ts
```diff
@@ -162,4 +171,44 @@ export default class PermissionsController extends BaseController {
...
+    const targetUser = await User.findOrFail(userId).addErrorContext(
+      () => `User with id ${userId} not found`,
+    );
+
+    const manager = Acl.model(targetUser);
+
+    for (const role of roles) {
+      await manager
+        .assignRole(role)
+        .addErrorContext(`Failed to assign role ${role}`);
```
mini-bomba: pass this error context via arrow function since it's a dynamically created string from a template (and not a static one)
also might as well include the user id
this comment applies to both new endpoints

app/controllers/v1/users.ts
```diff
@@ -0,0 +1,140 @@
...
+  async delete({ request, auth }: HttpContext) {
+    const user = await request.validateUsing(userIdParamValidator);
+
+    await this.requireSuperUser(auth);
+
+    const targetUser = await User.findOrFail(user.id).addErrorContext(
+      () => `User with id ${user.id} not found`,
+    );
+
+    await targetUser.delete();
```
mini-bomba: this should be wrapped in a transaction

mini-bomba (commented):
just some minor comments

mini-bomba (approved):
lgtm 🔥 :shipit:

PR #311, opened 2026-05-29
feat(mobile_config): add booths api base url & feature flag

### LINKED ISSUE #308: feat(mobile_config): booths api base url & feature flag
add a few fields to the mobile config model:
- one string that should be the base url of the booths api
- one boolean, to be used as a feature flag (nonnull, default false)

smigiska:
Added boothsApiBaseUrl and boothsEnabled to mobile_configs.
Created database migration and updated MobileConfig model.
CLOSES: #308

### DIFF
--- a/app/models/mobile_config.ts
+++ b/app/models/mobile_config.ts
@@ -49,9 +49,7 @@ export default class MobileConfig extends BaseModel {
   })
   declare boothsApiBaseUrl: string | null;
 
-  @typedColumn({
-    type: "boolean",
-  })
+  @typedColumn({ type: "boolean", hasDefault: true })
   declare boothsEnabled: boolean;
 
   @typedColumn.dateTime({ autoCreate: true })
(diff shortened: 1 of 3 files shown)


app/models/mobile_config.ts
```diff
@@ -40,6 +40,20 @@ export default class MobileConfig extends BaseModel {
...
+    validator: vine
+      .string()
+      .url({ protocols: ["https"], require_protocol: true })
+      .optional(),
+  })
+  declare boothsApiBaseUrl: string | null;
+
+  @typedColumn({
+    type: "boolean",
+  })
```
czaja307:
```suggestion
  @typedColumn({ type: "boolean", hasDefault: true })
```

Without this flag, the auto-generated CREATE validator will require `boothsEnabled` in requests, even though the database will provide a default.

czaja307 (changes requested):
one small suggestion - otherwise LGTM :fire:

czaja307 (approved):
🔥 :shipit:

PR #312, opened 2026-05-29
refactor: add transaction handling in auto CRUD controller

### LINKED ISSUE #277: chore: add transactions, everywhere
we've kinda neglected actually using transactions, we probably should start using them at some point
most of the changes will likely be in base controller, though there might be some extra changes needed elsewhere

krzys13:
#277 
Add transaction usage in all methods of auto crud controlle.
There is need to add transactions in 'user managment API'
Should I apply trnsactions in seeders, and scrappers also?


app/controllers/auto_crud_controller.ts
czaja307: `attach()` performs write operation on a pivot table - it should be wrapped in transaction

czaja307 (dismissed):
i found one more place that could use transactions - otherwise seems good :fire:

i don't thing we need transactions on seeders or scrappers as they are only used once without much concurrency risc

app/controllers/auto_crud_controller.ts
```diff
@@ -791,31 +810,49 @@ export default abstract class AutoCrudController<
...
+   */
+  protected async getFirstOrFail(
+    id: string | number,
+    trx?: TransactionClientContract,
+  ) {
     const primaryColumnName = this.primaryKeyField.columnOptions.columnName;
-    return await this.model
-      .query()
+    const data = await this.model
+      .query(trx !== undefined ? { client: trx } : undefined)
```
mini-bomba, line 838:
```suggestion
      .query(optionMap(trx, (trx) => ({ client: trx })))
```
(from `@solvro/utils/option`)

app/controllers/auto_crud_controller.ts
```diff
@@ -791,31 +810,49 @@ export default abstract class AutoCrudController<
...
+    if (trx !== undefined) {
+      row.useTransaction(trx);
+    }
+    const updatedRow = await row.save().addErrorContext({
       message: "Failed to commit updates",
       code: "E_DB_ERROR",
       status: 500,
     });
+    return updatedRow;
   }
```
mini-bomba, line 854: this change isn't really needed - if a transaction was used for fetching a model, the resulting instance will use that same transaction whenever any method call triggers an SQL query

app/controllers/auto_crud_controller.ts
mini-bomba, line 917: you should probably start the transaction when the model is first fetched

app/controllers/auto_crud_controller.ts
```diff
@@ -902,10 +948,13 @@ export default abstract class AutoCrudController<
...
-      message: "Failed to delete object",
-      code: "E_DB_ERROR",
-      status: 500,
+    await db.transaction(async (trx) => {
+      record.useTransaction(trx);
+      await record.delete().addErrorContext({
+        message: "Failed to delete object",
+        code: "E_DB_ERROR",
+        status: 500,
+      });
```
mini-bomba: same here - start the transaction when fetching the model

krzys13: done

app/controllers/auto_crud_controller.ts
krzys13: done

app/controllers/auto_crud_controller.ts
krzys13: done

app/controllers/auto_crud_controller.ts
krzys13: Test it, then tests fails about 11 of them. Suggestions did not apply.

mini-bomba: you need to import the func first, which i did not include in the suggestion snippet

krzys13:
We need to discuss 
1. Isolations levels fot querries
2. How to implement transactions in validators, Validators operate on http body{request} not on specific model instance.

mini-bomba:
for isolation levels, i'd suggest `repeatable read`

to implement transactions in the foreign key validator, i'd suggest passing in the `trx` object via [validation metadata](https://vinejs.dev/docs/schema_101#validation-metadata)

krzys13:
> for isolation levels, i'd suggest `repeatable read`
> 
> 
> 
> to implement transactions in the foreign key validator, i'd suggest passing in the `trx` object via [validation metadata](https://vinejs.dev/docs/schema_101#validation-metadata)

Isolations are done
Validation still not

krzys13:
I don't think this validator should exist here, because it only validates the existence of a foreign key.

This seems to already be handled by the `oneToOneRelationStore` method:

```ts
/**
 * Create a related object for 1:1 relations
 *
 * Return type set to Promise<unknown> to allow for method overrides
 */
async oneToOneRelationStore(httpCtx: HttpContext): Promise<unknown>
```

which handles creating the related object during the store flow.

@mini-bomba

mini-bomba:
I'm pretty sure I've made the validators for related-store operations omit the foreign key field on the related object to be created, so it shouldn't matter whether the field has a foreign key validator or not

the foreign key validator is actually executed in two scenarios:
- for direct store ops of objects with foreign keys
- for related-store ops of related objects with multiple foreign keys (validators for foreign key fields for other relations are still used)

(oh and i guess for updates too)

krzys13:
> I'm pretty sure I've made the validators for related-store operations omit the foreign key field on the related object to be created, so it shouldn't matter whether the field has a foreign key validator or not
> 
> the foreign key validator is actually executed in two scenarios:
> 
> * for direct store ops of objects with foreign keys
> * for related-store ops of related objects with multiple foreign keys (validators for foreign key fields for other relations are still used)
> 
> (oh and i guess for updates too)

I think we can close that issue and start another about validator refactor. Should checking fk be only in crud controller for better arch.

app/controllers/auto_crud_controller.ts
```diff
@@ -723,56 +736,94 @@ export default abstract class AutoCrudController<
...
+    httpCtx: HttpContext,
+    validator: AnyValidator,
+    trx: TransactionClientContract,
+  ): Promise<unknown> {
+    // Autogenerated validators are cached without metadata typing. This cast lets
+    // Vine pass the current transaction to custom rules through field.meta.
+    const transactionValidator = validator as unknown as VineValidator<
+      SchemaTypes,
+      { transaction: TransactionClientContract }
+    >;
```
mini-bomba: instead of casting through unknown, just change the definition of `AnyValidator` to accept the optional transaction metadata param

app/controllers/auto_crud_controller.ts
```diff
@@ -723,56 +736,94 @@ export default abstract class AutoCrudController<
...
+
     await this.authorizeRecord(httpCtx, "show", data);
     return { data };
   }
 
+  protected async validateUsingTransaction(
+    httpCtx: HttpContext,
+    validator: AnyValidator,
+    trx: TransactionClientContract,
+  ): Promise<unknown> {
```
mini-bomba: add a jsdoc comment for this function

mini-bomba (commented):
just some minor things at this point

sadly the diffs are nearly unreadable when indentation is modified, so i'll be assuming you did not accidentally change something you shouldn't have

app/controllers/auto_crud_controller.ts
```diff
@@ -723,56 +736,94 @@ export default abstract class AutoCrudController<
...
-      code: "E_DB_ERROR",
-      status: 500,
-    });
-
-    await result.refresh().addErrorContext({
-      message: "Failed to fetch updated object",
-      code: "E_DB_ERROR",
-      status: 500,
-    });
+    let toStore!: PartialModel<T>;
```
mini-bomba: instead of using a non-null assert (which i'm surprised wasn't picked up by our linter) execute the post-hook inside the transaction (possibly passing in the transaction as a parameter)

app/controllers/auto_crud_controller.ts
```diff
@@ -723,56 +736,94 @@ export default abstract class AutoCrudController<
...
+
+        const Model = this.model;
+        const createdModel = new Model() as InstanceType<T>;
+        createdModel.merge(toStore);
+        createdModel.useTransaction(trx);
+        await createdModel.save().addErrorContext({
+          message: "Failed to store object",
+          code: "E_DB_ERROR",
+          status: 500,
+        });
```
mini-bomba: `this.model.create()` accepts a 3rd `options` parameter, an object which may contain the `client` property which you should be able to set to the transaction handle

app/controllers/auto_crud_controller.ts
```diff
@@ -791,31 +842,49 @@ export default abstract class AutoCrudController<
...
+    if (trx !== undefined) {
+      row.useTransaction(trx);
+    }
+    const updatedRow = await row.save().addErrorContext({
       message: "Failed to commit updates",
       code: "E_DB_ERROR",
       status: 500,
     });
+    return updatedRow;
   }
```
mini-bomba, line 854: you forgot to revert this change after resolving the previous comment on this block

app/controllers/auto_crud_controller.ts
```diff
@@ -835,29 +904,42 @@ export default abstract class AutoCrudController<
...
+    const row = await db.transaction(
+      async (trx) => {
+        let updates = (await this.validateUsingTransaction(
+          httpCtx,
+          this.updateValidator,
+          trx,
+        )) as PartialModel<T>;
+
+        const searchedRow = await this.getFirstOrFail(id, trx);
+        await this.authorizeRecord(httpCtx, "update", searchedRow);
```
mini-bomba: authorization methods should likely accept a transaction parameter
`authorizeRecord` isn't actually implemented anywhere yet (but is there as an option should we need it later)
`authorizeById` is used in the drafts controller - i'd say it should also be executed in the same transaction, and should take a transaction param
here's a link to the library docs on how to use transactions with permission queries: https://www.npmjs.com/package/@holoyan/adonisjs-permissions#transactions

app/controllers/auto_crud_controller.ts
```diff
@@ -835,29 +904,42 @@ export default abstract class AutoCrudController<
...
+        updates =
+          (await this.updateHook({
+            http: httpCtx,
+            model: this.model,
+            record: searchedRow,
+            request: updates,
+          })) ?? updates;
+
+        searchedRow.merge(updates);
+        searchedRow.useTransaction(trx);
```
mini-bomba: unnecessary `.useTransaction` - model instances should inherit the query client from the fetch func

app/controllers/auto_crud_controller.ts
```diff
@@ -835,29 +904,42 @@ export default abstract class AutoCrudController<
...
+        const updatedRow = await searchedRow.save().addErrorContext({
+          message: "Failed to commit updates",
+          code: "E_DB_ERROR",
+          status: 500,
+        });
```
mini-bomba: why did you inline `this.saveOrFail`?
by my comment under the `saveOrFail` definition i meant that only the `.useTransaction` call should be done on the caller side (if necessary), not that the function should be inlined wherever transactions are used

app/controllers/auto_crud_controller.ts
```diff
@@ -902,11 +984,17 @@ export default abstract class AutoCrudController<
...
+      async (trx) => {
+        record.useTransaction(trx);
+        await record.delete().addErrorContext({
+          message: "Failed to delete object",
+          code: "E_DB_ERROR",
+          status: 500,
+        });
+      },
+      { isolationLevel: "repeatable read" },
+    );
```
mini-bomba, line 971: authorize, fetch and hook calls should likely be included in the transaction
hook definition should be modified to accept a transaction parameter

app/controllers/auto_crud_controller.ts
```diff
@@ -1002,65 +1096,73 @@ export default abstract class AutoCrudController<
...
+          trx,
+        )) as Partial<ModelAttributes<LucidRow>>;
+
+        const mainInstance = await this.getFirstOrFail(id, trx);
+        await this.authorizeRecord(
+          { request, route, auth } as unknown as HttpContext,
+          "oneToOneRelationStore",
+          mainInstance,
+        );
+        mainInstance.useTransaction(trx);
```
mini-bomba: likely unnecessary `.useTransaction()`

app/controllers/auto_crud_controller.ts
```diff
@@ -1091,104 +1193,125 @@ export default abstract class AutoCrudController<
...
+    const res = await db.transaction(
+      async (trx) => {
+        const toStore = (await this.validateUsingTransaction(
+          httpCtx,
+          this.relatedStoreValidator(relationName),
+          trx,
+        )) as Partial<ModelAttributes<LucidRow>>;
+
+        const mainInstance = await this.getFirstOrFail(id, trx);
+        mainInstance.useTransaction(trx);
```
mini-bomba: likely unnecessary `.useTransaction()`

app/controllers/auto_crud_controller.ts
```diff
@@ -1091,104 +1193,125 @@ export default abstract class AutoCrudController<
...
+        )) as Record<string, unknown>;
+
+        await this.authorizeById(httpCtx, "manyToManyRelationAttach", {
+          localId,
+          relatedId,
+          relationName,
+        });
+
+        const mainInstance = await this.getFirstOrFail(localId, trx);
+        mainInstance.useTransaction(trx);
```
mini-bomba: likely unnecessary `.useTransaction()`

app/controllers/auto_crud_controller.ts
```diff
@@ -1210,6 +1333,9 @@ export default abstract class AutoCrudController<
     const detachFilters = (await request.validateUsing(
       this.detachValidator(relationName),
     )) as Record<string, unknown>;
+    const definedDetachFilters = Object.fromEntries(
+      Object.entries(detachFilters).filter(([, value]) => value !== undefined),
+    );
```
mini-bomba: is this change actually needed?
vine should already be taking care of this, it should be removing any `undefined` values entirely (according to docs)

app/controllers/auto_crud_controller.ts
```diff
@@ -1233,40 +1359,55 @@ export default abstract class AutoCrudController<
...
+            status: 500,
+          });
+        }
 
-    if (result === 0) {
-      throw new NotFoundException("No relation attachments matched your query");
-    }
+        const result = Array.isArray(deleted)
+          ? deleted.length
+          : Number(deleted);
```
mini-bomba: is this addition actually needed? from my testing the return type there was always a number (since no `returning` clause was specified)

mini-bomba (changes requested):
i figured out how to get proper diffs locally

mini-bomba (approved):
:shipit:

PR #313, opened 2026-05-31
refactor(contributors): move order to pivot table

### LINKED ISSUE #314: feat: contributors order per milestone

czaja307:


mini-bomba:
there's a reason why we didn't add the order field to the pivot table
the reason being that the API would absolutely suck to use (changing the value would involve deleting and recreating the relationship)

i see two solutions:
- either make the pivot table an actual table & model (breaking change)
- or hack in a way to modify n:m relationship properties (which honestly shouldn't be *that* difficult, it would just be jank, as we'd be circumventing adonis' model system)

czaja307:
add transactions

simon-the-shark:
👀👀👀

app/utils/model_autogen.ts
```diff
@@ -359,6 +412,50 @@ export class AutogenCacheEntry {
...
+      continue;
+    }
+    if (field.detachFilter) {
+      pivotKey[name] = body[name];
+    } else {
+      pivotUpdate[name] = body[name];
+    }
+  }
+  return { pivotKey, pivotUpdate };
+}
```
mini-bomba: instead of filtering fields like this, the generated validator should have two objects: a `query`/`filter` object (for narrowing down the exact connection to modify) and an `update` object (with updates to apply)

mini-bomba (changes requested):

simon-the-shark:
👀👀👀👀👀👀

mini-bomba (approved):
🔥 :shipit:

PR #322, opened 2026-07-05
test: add tests for User Management API

### LINKED ISSUE #316: test: add tests for user management api

smigiska:
This PR completes the test suite for the User Management API. It also applies requested fixes to the controller validation. I added functional tests for all CRUD operations and error handling. The userIdParamValidator now correctly parses IDs from the URL. I also added a check for email uniqueness in the create and update methods. All tests passed successfully.
CLOSES: #316

### DIFF
--- a/app/controllers/v1/users.ts
+++ b/app/controllers/v1/users.ts
@@ -6,6 +6,7 @@ import type { Constructor, LazyImport } from "@adonisjs/core/types/http";
 import db from "@adonisjs/lucid/services/db";
 
 import User from "#app/models/user";
+import { ConflictException } from "#exceptions/http_exceptions";
 
 import BaseController from "../base_controller.js";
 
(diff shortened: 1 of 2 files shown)


app/controllers/v1/users.ts
```diff
@@ -62,46 +64,65 @@ export default class UsersController extends BaseController {
   }
 
   async findOne({ request, auth }: HttpContext) {
-    const user = await request.validateUsing(userIdParamValidator);
+    const { params } = await request.validateUsing(userIdParamValidator, {
+      data: { params: request.params() },
+    });
```
mini-bomba: that second parameter shouldn't be needed

app/controllers/v1/users.ts
```diff
@@ -62,46 +64,65 @@ export default class UsersController extends BaseController {
...
+    if (payload.email !== undefined) {
+      const existingUser = await User.query()
+        .where("email", payload.email)
+        .whereNot("id", params.id)
+        .first();
+
+      if (existingUser !== null) {
+        return response.unprocessableEntity({
+          error: "User with this email already exists",
+        });
```
mini-bomba: a 409 (conflict) return code would be better here
and also throw one of the exceptions from `app/exceptions/http_exceptions.ts` instead of manually returning an error to ensure the error response is consistent with all other controllers

also instead of manually checking whether a user with a given email exists before creating/editing the user, wrap the `await targetUser.save()` call in a try/catch block, and try to find out what kind of error you get in the catch section.
in the case of a duplicate email, you should get a postgresql error with a code property equal to `"23505"` - unique_violation (all possible error codes are listed in [appendix A of postgresql docs](https://www.postgresql.org/docs/18/errcodes-appendix.html))

tests/functional/users_management.spec.ts
```diff
@@ -0,0 +1,600 @@
...
+    await assignSolvroAdmin(admin);
+    const token = await makeToken(admin);
+    const fakeId = 999999;
+
+    const response = await client
+      .get(`/api/v1/users/${fakeId}`)
+      .bearerToken(token);
+
+    response.assertStatus(404);
+  });
```
mini-bomba, line 237: suggestion for more tests: you should test whether guests and authenticated non-admins get 401/403 on nonexistent user ids as well - otherwise we're leaking data to unauthorized users

tests/functional/users_management.spec.ts
```diff
@@ -0,0 +1,600 @@
...
+    const token = await makeToken(admin);
+
+    const fakeUserId = 999999;
+
+    const response = await client
+      .delete(`/api/v1/users/${fakeUserId}`)
+      .bearerToken(token);
+
+    response.assertStatus(404);
+  });
```
mini-bomba, line 482: same test suggestion as for `GET /api/v1/users/:id` - check whether guests and non-admins get 401/403 instead of 404 for nonexistent ids

tests/functional/users_management.spec.ts
```diff
@@ -0,0 +1,600 @@
...
+
+    const response = await client
+      .patch(`/api/v1/users/${fakeUserId}`)
+      .bearerToken(token)
+      .json({
+        fullName: "Guest Name",
+      });
+
+    response.assertStatus(404);
+  });
```
mini-bomba, line 617: same data leak test suggestion as for GET and DELETE

mini-bomba (commented):
nice work, but i do have some suggestions

mini-bomba (approved):
🔥 :shipit:

PR #324, opened 2026-07-13
feat: implement transaction handling in JWT and auto CRUD controllers

### LINKED ISSUE #315: feat(transactions): make auth functions execute in transactions
basically make `authenticate` functions in `AutoCrudController` take `trx` params and use these when querying the database

krzys13:
(#315)
- add trx usage in JWT auth
- edit autocrud controller
- add new config file for transaction isolation level

### DIFF
--- a/app/auth/guards/jwt.ts
+++ b/app/auth/guards/jwt.ts
@@ -68,7 +68,7 @@ interface RefreshTokenPayload extends SupportedPayload {
 
 export class JwtGuard implements GuardContract<User> {
   #ctx: HttpContext;
-  #trx?: TransactionClientContract;
+  #trx?: TransactionClientContract | undefined;
 
   constructor(ctx: HttpContext) {
     this.#ctx = ctx;
@@ -191,12 +191,8 @@ export class JwtGuard implements GuardContract<User> {
     return this;
   }
 
-  public async authenticate(
-    trx: TransactionClientContract | undefined = this.#trx,
-  ): Promise<User> {
-    this.#trx = undefined;
-
-    if (this.authenticationAttempted && trx === undefined) {
+  public async authenticate(): Promise<User> {
+    if (this.authenticationAttempted && this.#trx === undefined) {
       return this.getUserOrFail();
     }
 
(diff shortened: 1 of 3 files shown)


app/auth/guards/jwt.ts
```diff
@@ -184,14 +186,34 @@ export class JwtGuard implements GuardContract<User> {
...
 
-  public async authenticate(): Promise<User> {
-    if (this.authenticationAttempted) {
+  public useTransaction(trx: TransactionClientContract): this {
+    this.#trx = trx;
+    return this;
+  }
+
+  public async authenticate(
+    trx: TransactionClientContract | undefined = this.#trx,
```
mini-bomba: is this parameter needed if you're using the `useTransaction` method?

app/auth/guards/jwt.ts
```diff
@@ -184,14 +186,34 @@ export class JwtGuard implements GuardContract<User> {
...
+  ): Promise<User> {
+    this.#trx = undefined;
```
mini-bomba: is this needed?

krzys13: it is not, change applied

app/auth/guards/jwt.ts
krzys13: it is not, change applied

app/auth/guards/jwt.ts
```diff
@@ -184,14 +186,30 @@ export class JwtGuard implements GuardContract<User> {
...
+      this.authenticationAttempted = true;
+      const token = this.extractTokenFromHeaderOrFail();
+      const payload = this.validateAccessToken(token);
+      userId = payload.sub;
+    }
+
+    const owner = await User.find(
+      userId,
+      this.#trx === undefined ? undefined : { client: this.#trx },
+    );
```
mini-bomba, line 212: to discuss today: how to handle the issue of embedding the transaction in the user object that may be shared across auth calls

mini-bomba: in the short-circuit if case above instead of checking if `this.#trx` is none, check whether it is the same as the one embedded in `this.user`

mini-bomba (approved):
🔥 :shipit: 
(provided that i didn't break anything with my last commit)

PR #328, opened 2026-07-16
refactor: deduplicate unit test support functions (#323)

### LINKED ISSUE #323: refactor: deduplicate unit test support functions
move functions like `makeToken`, `uniqueEmail` or `ensureSolvroAdminRoleId` from unit test files to a dedicated, shared file

smigiska:
Extracted repeated helper functions (uniqueEmail, makeToken, ensureSolvroAdminRoleId, assignSolvroAdmin) from drafts_acl.spec.ts, permissions.spec.ts, permissions_cleanup.spec.ts, and users_management.spec.ts into a shared auth_helpers.ts module, imported wherever needed.
Also added createUserWithToken / createAdminWithToken helpers to reduce repeated "create user + get token" boilerplate across the same test files.
CLOSES: #323

### DIFF
--- a/tests/functional/auth_helpers.ts
+++ b/tests/functional/auth_helpers.ts
@@ -6,11 +6,9 @@ import db from "@adonisjs/lucid/services/db";
 import User from "#models/user";
 import env from "#start/env";
 
-export function createUniqueEmail(domain: string) {
-  return (prefix: string): string => {
-    const id = crypto.randomUUID().slice(0, 8);
-    return `${prefix}-${id}@${domain}`;
-  };
+export function uniqueEmail(prefix: string): string {
+  const id = crypto.randomUUID().slice(0, 8);
+  return `${prefix}-${id}@example.test`;
 }
 
 export async function makeToken(user: User): Promise<string> {
@@ -95,30 +93,30 @@ export async function assignSolvroAdmin(user: User) {
 }
 
 export async function createUserWithToken(
-  uniqueEmail: (prefix: string) => string,
   prefix: string,
   fullName: string,
-): Promise<{ user: User; token: string }> {
+): Promise<{ user: User; token: string; password: string }> {
+  const password = `Test-${crypto.randomUUID().slice(0, 8)}!`;
   const user = await User.create({
     email: uniqueEmail(prefix),
-    password: "Passw0rd!",
+    password,
     fullName,
   });
   const token = await makeToken(user);
-  return { user, token };
+  return { user, token, password };
 }
 
 export async function createAdminWithToken(
-  uniqueEmail: (prefix: string) => string,
   prefix: string,
   fullName: string,
-): Promise<{ user: User; token: string }> {
+): Promise<{ user: User; token: string; password: string }> {
+  const password = `Test-${crypto.randomUUID().slice(0, 8)}!`;
   const user = await User.create({
     email: uniqueEmail(prefix),
-    password: "Passw0rd!",
+    password,
     fullName,
   });
   await assignSolvroAdmin(user);
   const token = await makeToken(user);
-  return { user, token };
+  return { user, token, password };
 }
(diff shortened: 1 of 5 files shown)


tests/functional/auth_helpers.ts
```diff
@@ -0,0 +1,124 @@
...
+}
+
+export async function createUserWithToken(
+  uniqueEmail: (prefix: string) => string,
+  prefix: string,
+  fullName: string,
+): Promise<{ user: User; token: string }> {
+  const user = await User.create({
+    email: uniqueEmail(prefix),
+    password: "Passw0rd!",
```
mini-bomba: minor suggestion: maybe generate a random password and return it along with the user object & token?

tests/functional/drafts_acl.spec.ts
```diff
@@ -125,20 +31,18 @@ test.group("Drafts ACL (per-model and class-level)", (group) => {
...
-    const u2 = await User.create({
-      email: uniqueEmail("u2"),
-      password: "Passw0rd!",
-      fullName: "User 2",
-    });
+    const { user: u1, token: t1 } = await createUserWithToken(
+      uniqueEmail,
+      "u1",
+      "User 1",
+    );
```
mini-bomba: passing in a function that generates email addresses seems weird
imo we can drop the idea of using separate fake domains per test file - the emails contain a random string anyway, they shouldn't collide.

tests/functional/drafts_acl.spec.ts
```diff
@@ -310,12 +210,13 @@ test.group("Drafts ACL (per-model and class-level)", (group) => {
...
-      fullName: "User 8",
-    });
-    const token = await makeToken(user);
+    const { user, token } = await createUserWithToken(
+      uniqueEmail,
+      "user8",
+      "User 8",
+    );
+
+    await user.refresh();
```
mini-bomba: why is this refresh here? it should probably be a part of the utility function if the function returns a bad user object without this

mini-bomba (commented):
🔥, but i've got a few suggestions

tests/functional/drafts_acl.spec.ts
smigiska: honestly i wasn't sure why refresh() was there at first, so I looked into it a bit and i don't know if it is actually needed. MakeToken() only needs the user's id and that's already set right after the user is created. I removed it, but I'm not 100% sure if it's correct, so let me know if it should stay

mini-bomba: in general adonis does not automatically refresh objects after pushing them into the database, and so a `.refresh()` call immediately after inserting can make some sense if later code for example relies on default values set in the database
I think the best solution here is to simply move the `.refresh()` call into the helper, so that we don't get such surprises later when using these util functions in tests

tests/functional/auth_helpers.ts
```diff
@@ -95,30 +93,30 @@ export async function assignSolvroAdmin(user: User) {
 }
 
 export async function createUserWithToken(
-  uniqueEmail: (prefix: string) => string,
   prefix: string,
   fullName: string,
-): Promise<{ user: User; token: string }> {
+): Promise<{ user: User; token: string; password: string }> {
+  const password = `Test-${crypto.randomUUID().slice(0, 8)}!`;
```
mini-bomba, line 99: really minor issue: this is quite a weak password (doesn't *really* matter though, as these are only used in unit tests)
but if you'd like to actually generate somewhat-good passwords, I'd recommend you take a look at the [`crypto.randomBytes`](https://nodejs.org/docs/latest/api/crypto.html#cryptorandombytessize-callback) - generating something like 16 or 32 random bytes should be good enough, and then you can simply encode them into hex or base64 (with `.toString("hex/base64")` on the returned buffer)

(and if you do, then you might as well replace the `crypto.randomUUID().slice(0,8)` hack with a proper `crypto.randomBytes(4).toString("hex")`)

mini-bomba (approved):
:shipit:

PR #339, opened 2026-08-16
feat: implement Fields of Study scraper with Cheerio integration

### LINKED ISSUE #321: feat: field of studies scraper
Write a scraper that will scrape all majors from: [https://rekrutacja.pwr.edu.pl/](https://rekrutacja.pwr.edu.pl/)

krzys13:
#321


database/scrapers/departments.ts
```diff
@@ -43,39 +32,33 @@ interface DepartmentLinkDraft {
 export default class DepartmentsScraper extends BaseScraperModule {
   static name = "Departments";
   static description =
-    "Scrapes pwr departments, field of study and departments links data from directus";
-  static taskTitle = "Scrape departments, field of study and departments links";
+    "Scrapes pwr departments and departments links data from directus";
+  static taskTitle = "Scrape departmentsand departments links";
```
czaja307:
```suggestion
  static taskTitle = "Scrape departments and departments links";
```

database/scrapers/fields_of_study.ts
```diff
@@ -0,0 +1,225 @@
...
+    });
+    const fieldsOfStudyDetails =
+      result.fieldsOfStudyDetails as unknown as FieldOfStudyDetails[];
+
+    for (const details of fieldsOfStudyDetails) {
+      const responseDetails = await fetch(details.url);
+      const htmlDetails = await responseDetails.text();
+      const detailPage = cheerio.load(htmlDetails);
+
+      details.department = detailPage("div.course-header__top a > span")
```
czaja307: i'd think about creating const with this XPath somewhere on the begining of the for better maintainability

database/scrapers/fields_of_study.ts
```diff
@@ -0,0 +1,225 @@
...
+        .text()
+        .trim();
+
+      for (const wrapper of detailPage(
+        ".course-data-bar__data-wrapper",
```
czaja307: same as above

database/scrapers/fields_of_study.ts
```diff
@@ -0,0 +1,223 @@
...
+  private readonly basicSecondDegreeTermSpan = 3;
+  private readonly extendedSecondDegreeTermSpan = 4;
+  private readonly secondDegreeTermSpans = [
+    this.basicSecondDegreeTermSpan,
+    this.extendedSecondDegreeTermSpan,
+  ];
+  private extendInit(studyLevel: string) {
+    const body = this.basicInit.body as URLSearchParams;
+    body.set("s_level", studyLevel);
+    return Object.assign(this.basicInit, body);
```
mini-bomba: deep-clone the const object first, then perform modifications
...or just make a factory method to mint fresh RequestInit objects on demand, potentially with your body already properly set

commands/db_scrape.ts
```diff
@@ -154,8 +161,9 @@ export abstract class BaseScraperModule {
   protected async fetchDirectusJSON(
     url: string,
     item: string,
+    init?: RequestInit,
   ): Promise<SourceResponse<unknown>> {
-    const response = await this.fetchJSON(url, item);
+    const response = await this.fetchJSON(url, item, init);
     return assertResponseStructure(response, item);
   }
```
mini-bomba, line 167: directus shouldn't need this

database/scrapers/fields_of_study.ts
```diff
@@ -0,0 +1,223 @@
...
+    if (department === null) {
+      throw new Error(
+        `Department "${details.department}" not found for field of study "${details.name}."`,
+      );
+    }
+
+    fieldOfStudy.departmentId = department.id;
+    fieldOfStudy.name = details.name;
+    fieldOfStudy.url = details.url;
+    fieldOfStudy.isEnglish = details.language === "angielski";
```
mini-bomba, line 122: you should probably assert that the language is either english or polish, so that we don't get surprised by new german studies marked as polish

database/scrapers/fields_of_study.ts
```diff
@@ -0,0 +1,223 @@
...
+    if (
+      this.firstDegreeTermSpans.some((value) =>
+        details.timeSpan.includes(value.toString()),
+      )
+    ) {
```
mini-bomba: what is the `details.timeSpan` value?
is it always exactly one of the expected values?
does it include leading/trailing whitespace?
can we get away with matching for this value without `.includes()`, using maybe `===` instead?

database/scrapers/fields_of_study.ts
```diff
@@ -0,0 +1,223 @@
...
+        if (existingFieldOfStudy !== null) {
+          existingFieldOfStudy.merge(fieldOfStudy.$attributes);
+          await existingFieldOfStudy.save();
+        } else {
+          await fieldOfStudy.save();
+        }
+      } catch (error) {
+        this.logger.warning(
+          `${error} Insertion of this field of study skipped`,
+        );
```
mini-bomba, line 228: add a context sentence before the error, like "Failed to write the '{name}' field of study to database: "
consider using `prepareReportForLogging` to generate a full error report, since the error is not controlled by you

mini-bomba (changes requested):

database/scrapers/fields_of_study.ts
krzys13: There is only  `is_english flag` in `FieldOfStudyModel` . In "german case' there will be just `is_english` set to 0. Should i edit language atributes of this model?

database/scrapers/fields_of_study.ts
krzys13: It is a statment , ex.  '3 semestry', ' 7 semestrów'.. . So i need to extract number from this

database/scrapers/fields_of_study.ts
mini-bomba, line 1: add some extra status messages so that i'm not left wondering whether the script is doing something or not when it takes over a minute to run

mini-bomba (changes requested):

mini-bomba:
image

uhhhh

mini-bomba (changes requested):
I've applied the changes I wanted to make the progress tracker myself, and I've fixed the issue with running commands on node 26.

But while testing I found that we're silently dropping fields of study, despite them being accounted for in the progress tracker.

I've counted the number of available fields on the recruitment page: 64+67 = 131
Image

The script prints no warnings and claims that all fields were scraped
Image

Despite this, the database after scraping only contains 92 entries, missing 39 fields of study
Image

krzys13:
> I've applied the changes I wanted to make the progress tracker myself, and I've fixed the issue with running commands on node 26.
> 
> But while testing I found that we're silently dropping fields of study, despite them being accounted for in the progress tracker.
> 
> I've counted the number of available fields on the recruitment page: 64+67 = 131 Image
> 
> The script prints no warnings and claims that all fields were scraped Image
> 
> Despite this, the database after scraping only contains 92 entries, missing 39 fields of study Image

During writing fetched models to database existing overwriting issue. When script look for exsiting field of studies it serach only by name. In cause of field of studies with 1DEGREE and 2DEGREE , 1DEGREE wasa indetified as same as 2DEGREE so two diffrent rows was written  as one by merege.
RESOLVED
In

mini-bomba (approved):
🔥 :shipit:

PR #340, opened 2026-08-17
feat(mobile_config): add killswitchOfDoomAndDespair flag

### LINKED ISSUE #320: feat: app kill switch
Add a kill switch to add in-app banner in case of EOL

czaja307:
i hope that we can keep this field name :D
closes #320


mini-bomba (approved):
🔥 :shipit: 
(can't wait for this to break prod due to adonis converting the column name between camelcase and snakecase differently)

PR #349, opened 2026-09-19
refactor: change applyInhertinace method

krzys13:
Change applyInhertinace method to change iteration scope. In previouis version iteration on fist verion of api was useless.


app/utils/controllers.ts
```diff
@@ -151,10 +151,13 @@ export async function importAllControllers(): Promise<ImportedApiVersion[]> {
...
   // sort the versions ascending
   listing.sort((a, b) => a.apiVersion - b.apiVersion);
+  //first version's controllers
   let previousVersion = listing[0];
+  //newer version's controllers
+  const nextVersions = listing.slice(1);
 
   // iterate over the listing
-  for (const version of listing) {
+  for (const version of nextVersions) {
```
mini-bomba: minor nitpick: might as well replace `listing[0]` with `listing.shift()`

krzys13: I tried your suggestions. It results in errors. Agter the editing listingi by `.shift()` method it also used during registering routerts and here throws errors. 
I can use a copy of this array

